PR #6, opened 2015-02-10
Allow same action name on diffrent ActionClasses

quazzie:
without this there can be for example only one action named 'update'.
with this you can have multiple ActionClasses with same actions.

class UserActions {
	constructor() { this.generateActions('update'); }
}
module.exports = alt.createActions(UserActions);

class LinkActions {
	constructor() { this.generateActions('update'); }
}
module.exports = alt.createActions(LinkActions);


goatslacker:
Thanks! I had changed this during a refactoring.

PR #17, opened 2015-02-22
removed flux npm module from chat example package.json

jbhatab:


goatslacker:
Can you add `alt` also? 

`"alt": "^0.13.5"`

jbhatab:
well isn't it self referencing? it references the repo for alt. Do you want me to change the require call to alt too?

goatslacker:
No because examples don't get checked-in to npm.

On changing the requires, sure.

Here's the npm package for this: https://www.npmjs.com/package/alt-chat

goatslacker:
Thanks!

PR #20, opened 2015-02-23
Adding a shorthand form of alt.createActions()

1000hz:
If you're not opposed to overloading the API, I think it could be useful to implement a shorthand for `alt.createActions()` that accepts an array of action names and automatically calls `this.generateActions()` for you in the constructor.

This pull request makes it so that the following two snippets are equivalent:

```javascript
var mySimpleActions = alt.createActions(['actionOne', 'actionTwo', 'actionThree'])
```
---
```javascript
class MySimpleActions {
  constructor() {
    this.generateActions('actionOne', 'actionTwo', 'actionThree')
  }
}

var mySimpleActions = alt.createActions(MySimpleActions)
```

Something to consider is that there would be an inconsistency between `generateActions` accepting multiple arguments and `createActions` accepting an array. Using a spread operator on the first argument of `createActions`  would be a little tricky because of the `exportObj` argument. Let me know your thoughts.

### DIFF
--- a/dist/alt.js
+++ b/dist/alt.js
@@ -11,6 +11,7 @@ var _prototypeProperties = function (child, staticProps, instanceProps) { if (st
 var _classCallCheck = function (instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } };
 
 var Dispatcher = require("flux").Dispatcher;
+
 var EventEmitter = _interopRequire(require("eventemitter3"));
 
 var Symbol = _interopRequire(require("es-symbol"));
@@ -58,6 +59,7 @@ var getInternalMethods = function (obj, excluded) {
 var AltStore = (function () {
   function AltStore(dispatcher, state) {
     var _this5 = this;
+
     _classCallCheck(this, AltStore);
 
     this[EE] = new EventEmitter();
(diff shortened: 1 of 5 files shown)


goatslacker:
It's a good idea provided it's small enough. Let's rename it to generateActions and make it a splat and this is :+1:

goatslacker:
Oh you'll need to add tests as well. Please

1000hz:
The reason I didn't use a splat originally was that with the default parameter on `exportObj` there wouldn't be a way for the interpreter to discern where to put the last argument. Is that not the case?

1000hz:
unless you mean splat them all into `...args` and do type checking ourselves.

goatslacker:
Right makes sense, sorry I didn't make sense before, let's split this out into a new function called generateActions which works just like generateActions.

1000hz:
Alright, see if this looks good to you.

goatslacker:
Sweet. Last step, can you run `npm test` or `npm run build` and then check that in. It'll run the build step.

1000hz:
:tada: :confetti_ball: :tada:

dist/alt.js
```diff
@@ -58,6 +59,7 @@ var getInternalMethods = function (obj, excluded) {
 var AltStore = (function () {
   function AltStore(dispatcher, state) {
     var _this5 = this;
+
```
goatslacker: lol I don't even know.

goatslacker:
Thanks for the PR! :100:

PR #22, opened 2015-02-24
Fixing breaking change that prevented server side rendering when ListenerMixin was used.

cdebotton:
Changing lifecycle hook for Listener Mixin to Component Will Mount so that it functions identically between server rendering and client rendering.

### DIFF
--- a/dist/alt-with-runtime.js
+++ b/dist/alt-with-runtime.js
@@ -1,10 +1,9 @@
 "use strict";
 
 var Dispatcher = require("flux").Dispatcher;
+
 var EventEmitter = babelHelpers.interopRequire(require("eventemitter3"));
-
 var Symbol = babelHelpers.interopRequire(require("es-symbol"));
-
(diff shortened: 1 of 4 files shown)


mixins/ListenerMixin.js
```diff
@@ -1,7 +1,7 @@
 var Subscribe = require('./Subscribe')
 
 var ListenerMixin = {
-  componentDidMount: function () {
+  componentWillMount: function () {
```
goatslacker: Thanks!

dist/alt-with-runtime.js
```diff
@@ -1,10 +1,9 @@
 "use strict";
 
 var Dispatcher = require("flux").Dispatcher;
-var EventEmitter = babelHelpers.interopRequire(require("eventemitter3"));
```
goatslacker: can you revert this change?

goatslacker:
I cherry picked this change here: a3a83b963c970e44db10f13afd0c20f74d01084b

Thank you!

PR #41, opened 2015-03-06
Extract function name from source as a fallback to support IE

philippotto:
Unfortunately, Internet Explorer (I know, I know...) doesn't support `Function.prototype.name`, yet (and I think other browsers are lacking it, as well. See [here](http://kangax.github.io/compat-table/es6/#function_name_property)). I can understand if you don't want to support older technologies and want to delegate this work to transpilers and polyfills. But in this case, it was quite an easy fix and now it works. So, I think it should be worth considering to merge this :)

I reordered the priority of the potential name sources a little bit. I changed `StoreModel.displayName` and `StoreModel.name` which made it more clean to refactor the extraction of the function name into an own function. Additionally, it makes more sense to prioritize the optional `displayName` property over the standard function name, I think.

By the way: The chat example works in IE because store names were explicitly passed to `alt.createStore` and the TodoMVC example worked because there is only one store and `undefined` is taken as the store name. As soon as you add another store, it will also get `undefined` as a name which breaks alt.

### DIFF
--- a/dist/alt-with-runtime.js
+++ b/dist/alt-with-runtime.js
@@ -27,6 +27,21 @@ var formatAsConstant = function (name) {
   }).toUpperCase();
 };
 
+var getFunctionName = function (aFunction) {
+  // Function.prototype.name is available
+  if (aFunction.name) {
+    return aFunction.name;
+  }
+
+  // Extract the function name from its source
+  var match = aFunction.toString().match(/^function\s*([^\s(]+)/);
+  if (match) {
+    return match[1];
+  }
+
+  return undefined;
+};
+
 /* istanbul ignore next */
 function NoopClass() {}
 
@@ -286,7 +301,7 @@ var Alt = (function () {
         var saveStore = arguments[2] === undefined ? true : arguments[2];
 
         var storeInstance = undefined;
-        var key = iden || StoreModel.name || StoreModel.displayName || "";
+        var key = iden || StoreModel.displayName || getFunctionName(StoreModel) || "";
 
         if (saveStore && (this.stores[key] || !key)) {
           /* istanbul ignore else */
--- a/dist/alt.js
+++ b/dist/alt.js
@@ -39,6 +39,21 @@ var formatAsConstant = function (name) {
   }).toUpperCase();
 };
 
+var getFunctionName = function (aFunction) {
+  // Function.prototype.name is available
+  if (aFunction.name) {
+    return aFunction.name;
+  }
+
+  // Extract the function name from its source
+  var match = aFunction.toString().match(/^function\s*([^\s(]+)/);
+  if (match) {
+    return match[1];
+  }
+
+  return undefined;
+};
+
 /* istanbul ignore next */
 function NoopClass() {}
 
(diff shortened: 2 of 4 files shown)


jdlehman:
Just want to weigh in my thoughts on adding specific browser support in general.  I like that Alt tries to be minimal and lightweight in general, so I like evaluating features as well as support very critically (we shouldn't be adding features for the sake of adding features unless there is a real need etc).

Since Alt is transpiled to ES5 with [Babel](https://babeljs.io/), a lot of the browser support will be achieved just from Babel, which is a huge win (as long as those browsers support the ES5 spec).  I think browser support should fall onto the user rather than Alt to implement.  For example, if ES3 support is needed, then the ES5 shim should be added by the user, not Alt.

Although this is a small change, I think it still adds unnecessary complexity to the library.

TLDR; Browser support should be the responsibility of the user, not Alt, to provide the needed shims and polyfills to run in the environments they desire.

troutowicz:
I agree with @jdlehman. The tweak is small, but it feels like too big of a change for alt.

philippotto:
I can understand your point, but when browser support should be the responsibility of the user, why does alt use babel at all? alt could be provided as pure ES6 and the user would have to take care of making it compatible.

Currently, I see that it is transpiled to ES5 compatible code with the exception that `Function.prototype.name` is broken because BabelJS doesn't take care of it. My proposed solution is only a few lines long and would make sure that it fully works on ES5 browsers (which seems to be the current scope of the project; otherwise babel wouldn't be needed in the build process). Or am I missing something?

jdlehman:
I see what you are getting at, but I think this is something that should be filed as an issue with [Babel](https://github.com/babel/babel/issues) instead because this falls under Babel's responsibility. Alt uses Babel to transpile to ES5, which by nature of using a dependency, brings along any bugs/problems that the dependency in question has, in this case `Function.prototype.name` is broken in IE.

philippotto:
Yeah, after having thought a bit about this, I also think that babel should
handle this. I will investigate the options and write back when I got a
solution for it.

Jonathan Lehman <notifications@github.com> schrieb am Fr., 6. März 2015
18:55:

goatslacker:
This was "fixed" in 48c589f5dfd5e623a3c6ab5b490a44ef319fc9ad

I'll have more details in a bit.

goatslacker:
Ok so to elaborate a bit more.

This isn't an issue with babel itself since the function name depends on the developer not on alt. That commit I referenced changed the behavior. Previously an error was thrown when a store was encountered with the same name.

Due to minification in production environments this can lead to unexpected bugs where it works on dev because some devs don't run the same minified stack on dev but then fails on prod because the source is minified and store names collide. For that reason I introduced the change to just console.warn whenever it is encountered.

If you plan on using snapshots it is recommended you pass in the store name yourself and do not rely on the magic. If you don't care however, feel free to continue using it and it won't throw any more errors.

There's also the saveStore parameter you can pass to createStore where it tells alt not to save that store for snapshots later. That'll also disable the check.

On "older browsers" Alt is supposed to work on IE8+ a few fixes landed in recently to make sure that IE8 support is more robust and errs on the side of safe handling.

goatslacker:
Thanks for the pull request and for the discussion! :)

Issue #58, opened 2015-03-09
Store class inheritance


quangbuule:
The Store class inheritance doesn't work:

```js
class BaseStore {
  static doSomethingOnBase() {
    console.log('doSomethingOnBase()');
    this.emitChange();
  }
};

class Store extends BaseStore {
  static doSomething() {
    console.log('doSomething()');
    this.emitChange();
  }
}

var storeInstance = alt.createStore(Store);

storeInstance.listen(() => {
  console.log('Triggered.');
});

console.log(BaseStore.doSomethingOnBase);

storeInstance.doSomething();
storeInstance.doSomethingOnBase();
```

`BaseStore.doSomethingOnBase` still exists but `storeInstance.doSomethingOnBase` is undefined.

jdlehman:
This does not work because of `Object.getOwnPropertyNames(obj)` works, which is being used [here](https://github.com/goatslacker/alt/blob/master/src/alt.js#L40).

See this [example](https://babeljs.io/repl/#?experimental=true&playground=false&evaluate=true&loose=false&spec=false&code=class%20Parent%20%7B%0A%20%20parentMethod()%20%7B%7D%0A%20%20static%20parentStaticMethod()%20%7B%7D%0A%7D%0A%0Aclass%20Child%20extends%20Parent%20%7B%0A%20%20childMethod()%20%7B%7D%0A%20%20static%20childStaticMethod()%20%7B%7D%0A%7D%0A%0A%2F%2F%20parent%0Aconsole.log(Object.getOwnPropertyNames(Parent))%3B%0A%2F%2F%20child%0Aconsole.log(Object.getOwnPropertyNames(Child))%3B) of what happens in Babel's repl.

I am also curious if extending a store is a common use case.  I cannot think of an example where this would be done in an application (that doesn't mean there isn't a use case, I just can't think of one off the top of my head).

troutowicz:
If there ends up being a use case for extending a store, this change should make that possible.

```javascript
let storeProto = {};
if (Object.getPrototypeOf(StoreModel) !== Object.getPrototypeOf(NoopClass)) {
  storeProto = getInternalMethods(Object.getPrototypeOf(StoreModel), builtIns));
}

storeInstance = assign(new AltStore(this.dispatcher, store), getInternalMethods(StoreModel, builtIns), storeProto);
```

quangbuule:
Inheritance is the beautiful thing come to ES6 :)

jdlehman:
[...]

PR #60, opened 2015-03-09
Inherit static methods of ancestor stores

troutowicz:
See issue #58

Just submitting this for review right now, I will write up some tests later. I don't like how many LOC this requires, maybe someone can find a good way to reduce. :)

### DIFF
--- a/dist/alt-with-runtime.js
+++ b/dist/alt-with-runtime.js
@@ -44,6 +44,23 @@ var getInternalMethods = function (obj, excluded) {
   }, {});
 };
 
+var getInheritedMethods = function (obj) {
+  var methods = arguments[1] === undefined ? {} : arguments[1];
+
+  if (Object.getPrototypeOf(obj) !== Object.getPrototypeOf(NoopClass)) {
+    var ancestorObj = Object.getPrototypeOf(obj);
+    var ancestorMethods = getInternalMethods(ancestorObj, builtIns);
+
+    for (var m in ancestorMethods) {
+      methods[m] = methods[m] || ancestorMethods[m];
+    }
+
+    return getInheritedMethods(ancestorObj, methods);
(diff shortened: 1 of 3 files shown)


dist/alt-with-runtime.js
```diff
@@ -44,6 +44,23 @@ var getInternalMethods = function (obj, excluded) {
   }, {});
 };
 
+var getInheritedMethods = function (obj) {
+  var methods = arguments[1] === undefined ? {} : arguments[1];
```
jdlehman: I would do

`var getInheritedMethods = function(obj, methods={}) { ... }` instead of using `var methods = arguments[1] === undefined ? {} : arguments[1];`.

jdlehman:
I'll take some time to see if I can reduce this at all later today.  Other than my one comment in the code, I think you need to rebase with the latest master to see alt's changes in coding style.  You can also check out the [contributing guide](https://github.com/goatslacker/alt/blob/master/CONTRIBUTING.md#coding-style).

I have not yet looked to see if this works, but I am sure your tests will help validate that :)

troutowicz:
@jdlehman 
You are looking at one of the built dist files, my changes originate in src/alt.js

jdlehman:
Gah, sorry about that! Makes way more sense now. I should not code review when tired lol.

goatslacker:
This needs tests and docs, but I'm not sold on it just yet. What happens when there are methods of confliciting names? Can you show some examples where this might be useful?

troutowicz:
As you go deeper into the ancestry, methods with a name already found in a child are skipped. I personally have no useful example to offer, and I agree this is almost too much. But it's worth a discussion nonetheless.

troutowicz:
Closed in favor of PR #66

PR #63, opened 2015-03-09
Edit store testing docs

jdlehman:
- Remove example that relies too much on Alt's inner-workings
- Clean up some wording


troutowicz:
:'(

I still don't like the idea of having to do a separate export solely for the ability to test. But the majority rules.

jdlehman:
Maybe we can build out some test utils to make this easier?

troutowicz:
What about if I alter my approach to find the symbol by string literal instead of index.

```javascript
const STATE_CONTAINER = Symbol('the state container')
```

Find the property that contains 'the state container'

goatslacker:
That's clever but the string literal can also change in the future :)

troutowicz:
Ah, I couldn't think of a reason why that would ever need to change. Merge away :)

goatslacker:
It probably wouldn't, but if it does change I wouldn't classify it as a breaking change.

PR #64, opened 2015-03-10
Add action testing documentation

jdlehman:


jdlehman:
Don't merge this yet, I am going to simplify this example later today to make it easier to understand.

Also, I think the last thing we can do is add some tests to the example alt apps and reference them in the testing docs. That way people can see a real example instead of something contrived.

jdlehman:
Ready for feedback, I updated the example so it is less confusing and more direct to the point. (no made up libraries doing magic!)

troutowicz:
@jdlehman in case you didn't know, coverage is at 100%

jdlehman:
@troutowicz I think another 20 notifications and I will really get it.

goatslacker:
good job summoning it again

goatslacker:
This looks nice. Let me load it up and give it a spin.

goatslacker:
Looks great :+1:

PR #66, opened 2015-03-10
exportPublicMethods mixin

troutowicz:
Issue #58

### DIFF
--- a/docs/createStore.md
+++ b/docs/createStore.md
@@ -120,6 +120,30 @@ class MyStore {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


jdlehman:
Looks good to me. The docs you added make the usage very clear as well.

goatslacker:
LGTM, how about an API of:

```js
this.exportPublicMethods({
  foo: this.foo
});
```

The key is what we'll use to export, the value is a reference to the function.

troutowicz:
Great idea @goatslacker, I'll modify tonight.

PR #68, opened 2015-03-10
Update test store docs so example works

jdlehman:
- User should be able to run this and see passing tests rather than having "approximate guidelines"

### DIFF
--- a/docs/testing/stores.md
+++ b/docs/testing/stores.md
@@ -129,13 +129,15 @@ describe('PetStore', () => {
   // though we can see that this method is working from our tests above,
   // lets use this inaccessible method to show how we can test
   // non static methods if we desire/need to
-  it('rounds money to 2 decimal places', function() {
+  it('rounds money to 2 decimal places', () => {
+    // we extend UnwrappedPetStore so we can easily stub out
+    // functions that alt adds to the wrapped store like `bindActions`
     class TestPetStore extends UnwrappedPetStore {
       constructor() {
-        // we extend and override the constructor because
-        // this.bindActions will fail in the unwrapped PetStore as
-        // it is an all specific method that the wrapped PetStore has
+        super();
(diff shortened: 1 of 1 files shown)


docs/testing/stores.md
```diff
@@ -113,7 +130,14 @@ describe('PetStore', () => {
   // lets use this inaccessible method to show how we can test
   // non static methods if we desire/need to
   it('rounds money to 2 decimal places', function() {
-    var unwrappedStore = new UnwrappedPetStore();
+    class TestPetStore extends UnwrappedPetStore {
+      constructor() {
+        // we extend and override the constructor because
```
goatslacker: This won't work with proper ES6 because derived classes need to call super()

jdlehman: Ah was not aware of that. We could just promote blackbox testing.  Any "private" functions should have affects that can be measured elsewhere in the store.  I know at least in the Ruby community, the testing of private methods is not endorsed.

So I am wondering if there is a case where we are unable to test the store just using its publicly accessible functions.  I actually think this might be rare given the purpose of stores.

jdlehman: Actually I am going to go ahead and mock `bindActions`. Though most testing might be able to be achieved through blackbox testing, I want to provide examples for users that wish to test the class directly or have a use case where blackbox testing is not enough.

goatslacker:
Are you sure @coveralls

goatslacker:
Cool, so the bindActions part is working now. But the store still doesn't export correctly...can you post what the transpiled code is?

jdlehman:
I can do you one better.  Here is a [repo](https://github.com/jdlehman/alt-example-tests) with working tests for you to play with :smile_cat:

goatslacker:
:+1: Thanks for this!

PR #70, opened 2015-03-11
Update bootstrap docs

jdlehman:
- Docs reference that bootstrap can only be done once, which is no  longer true thanks to this commit 0ba7b85
- Fix a few spelling mistakes in README

### DIFF
--- a/docs/bootstrap.md
+++ b/docs/bootstrap.md
@@ -11,6 +11,18 @@ permalink: /docs/bootstrap/
 
 The `alt.bootstrap()` function takes in a snapshot you've saved and reloads all the state with that snapshot, no events will be emitted to your components during this process, so it is advised to do this on init before the view has even rendered.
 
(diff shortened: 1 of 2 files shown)


docs/bootstrap.md
```diff
@@ -11,6 +11,18 @@ permalink: /docs/bootstrap/
 
 The `alt.bootstrap()` function takes in a snapshot you've saved and reloads all the state with that snapshot, no events will be emitted to your components during this process, so it is advised to do this on init before the view has even rendered.
 
+If you need to emit a change event, you can use `this.emitChange` inside of your `bootstrap` life cycle method.
```
goatslacker: If you want to emit a change event after bootstrapping you can use [`this.emitChange`](link to emit change) inside of the [bootstrap lifecycle method](lifecycle.md).

goatslacker:
Nice, this lgtm. Thanks!

docs/bootstrap.md
jdlehman: Updated

PR #75, opened 2015-03-13
Add deserialize lifecycle hook

jdlehman:
- `deserialize` is called before bootstrap/snapshot data is set back to the store. this allows the data to be transformed into a format that  the store expects. the return data becomes the store's state.
- `snapshot` now allows an optional return value. this value is used as the snapshot data if provided, otherwise the default `Store#getState()` is used.
- see the docs for more information

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -11,13 +11,13 @@ The [snapshot](lifecycleListeners.md#snapshot) and [deserialize](lifecycleListen
 
 In the example below, we will show how this technique can be used to `serialize` complex model data being used by the store into a simpler structure for the store's snapshot, and repopulate our store models with the simple structure from the bootstrap/snapshot data with `deserialize`.
 
-## Snapshot
+## Serialize
 
 `snapshot` provides a hook to transform the store data (via an optional return value) to be saved to an alt snapshot. If a return value is provided than it becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `serialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`. If there is no return value, the default, [`MyStore#getState()`](stores.md#storegetstate) is used for the snapshot data.
 
 ## Deserialize
 
-`deserialize` provides a hook to transform snapshot/bootstrap data into a form acceptable to the store for use within the application. The return value of this function becomes the state of the store. For example, if the `deserialize` received the data `{queryParams: 'val=2&val2=23'}`, we might transform the data into `{val: 2, val2: 23}` and return it to set the store data such that `myStore.val === 2` and `myStore.val2 === 23`. Deserialize can be useful for converting data from an external source such as a JSON API into the format the store expects.
+`deserialize` provides a hook to transform snapshot/bootstrap data into a form acceptable to the store for use within the application. The return value of this function becomes the state of the store. If there is no return value, the state of the store from the snapshot is used verbatim. For example, if the `deserialize` received the data `{queryParams: 'val=2&val2=23'}`, we might transform the data into `{val: 2, val2: 23}` and return it to set the store data such that `myStore.val === 2` and `myStore.val2 === 23`. Deserialize can be useful for converting data from an external source such as a JSON API into the format the store expects.
 
 ## Example
 
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -7,13 +7,13 @@ permalink: /docs/serialization/
 
 # Serialization
 
-The [serialize](lifecycleListeners.md#serialize) and [deserialize](lifecycleListeners.md#deserialize) lifecycle listener methods can be utilized separately or together to transform the store data being saved in a snapshot or the snapshot/bootstrap data being set to a store. Though they do not have to be used together, you can picture the parity between the two methods. You can transform the shape of your store data created in the snapshot with `serialize`, which might be to get your data in a format ready to be consumed by other services and use `deserialize` to transform this data back into the format that your store recognizes.
+The [snapshot](lifecycleListeners.md#snapshot) and [deserialize](lifecycleListeners.md#deserialize) lifecycle listener methods can be utilized separately or together to transform the store data being saved in a snapshot or the snapshot/bootstrap data being set to a store. Though they do not have to be used together, you can picture the parity between the two methods. You can transform the shape of your store data created in the snapshot with `snapshot`, which might be to get your data in a format ready to be consumed by other services and use `deserialize` to transform this data back into the format that your store recognizes.
 
 In the example below, we will show how this technique can be used to `serialize` complex model data being used by the store into a simpler structure for the store's snapshot, and repopulate our store models with the simple structure from the bootstrap/snapshot data with `deserialize`.
 
-## Serialize
+## Snapshot
 
-`serialize` provides a hook to transform the store data to be saved to an alt snapshot, application data persisted by alt. The return value of this function becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `serialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`
+`snapshot` provides a hook to transform the store data (via an optional return value) to be saved to an alt snapshot. If a return value is provided than it becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `serialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`. If there is no return value, the default, [`MyStore#getState()`](stores.md#storegetstate) is used for the snapshot data.
 
 ## Deserialize
 
@@ -58,7 +58,7 @@ class MyStore {
     // we don't want to save this data in the snapshot
     this.semiPrivateVal = 10
 
-    this.on('serialize', () => {
+    this.on('snapshot', () => {
       return {
         // provides product and sum data from the model getters in addition to x and y
         // this data would not be included by the default serialization (getState)
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -1,0 +1,86 @@
+---
+layout: docs
+title: Serialization
+description: Serialization example
+permalink: /docs/serialization/
+---
+
+# Serialization
+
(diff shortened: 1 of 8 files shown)


goatslacker:
We already have [snapshot](https://github.com/goatslacker/alt/blob/master/docs/lifecycleListeners.md#snapshot) which does what serialize would do.

And [bootstrap](https://github.com/goatslacker/alt/blob/master/docs/lifecycleListeners.md#bootstrap) which somewhat does what deserialize does.

What are these methods adding?

goatslacker:
Also, this is a badass pull request. Code, build, test, docs. Amazing!

jdlehman:
Both `snapshot` and `bootstrap` give you access around these areas, but do not do the same thing as `serialize` and `deserialize`.  Let me highlight the differences.

The `snapshot` lifecycle method is [called](https://github.com/goatslacker/alt/blob/master/src/alt.js#L228) right before the value for that store within the snapshot is [set](https://github.com/goatslacker/alt/blob/master/src/alt.js#L230).  This means that you could modify data in your store within the snapshot, but it does not give you full control over what is saved in the snapshot. That is what `getState` ends up doing.  I think `snapshot` is for ensuring everything is in order before the snapshot is taken, but it does not give you any power over the data that you save.  This is the power of `serialize`.  It lets you specify exactly what gets saved in that store's snapshot as the return value is the data that gets stored for that store.  So I think that `getState` (in the way it is being used for `snapshot`) is actually more similar to `serialize` than the `snapshot` lifecycle method is.  I don't think using `getState` is great to accomplish what `serialize` does, because `getState`'s job is to provide data to the view, whereas `serialize` specifies what data gets saved in a snapshot. The data that goes to the view might actually be slightly different, in structure or content, than what you want to take a snapshot of.

`deserialize` is `serialize`'s trusty companion (and the more powerful of the two in my opinion), as it does the inverse of `serialize` in a sense. `bootstrap` is called after the data from the `snapshot` has already been set on the store.  This means that if you want to use `bootstrap` to reformat data, you have to deal with the fact that the data has already been set on the store exactly how it was in the snapshot.  I tried this approach initially and it led me to seek something better. `deserialize` receives the data right before it is set to the store, and expects the return of the function to be the data in the format you want to save it to the store.  This means that I don't have to do any cleanup of the store (removing properties I didn't want set, etc) since `deserialize` hooks in directly.

I think the biggest difference and power in `serialize` and `deserialize` is that they are closer to the metal.  They hook directly into how the snapshot/bootstrap data is serialized/deserialized, which provides HUGE flexibility, and could even aid future enhancements to alt.

The need I am directly filling right now is that my company's app's stores are backed by a variety of models that provide immutable helpers and various computed properties in the form of getters.  Some of these models extend each other as they share a lot of functionality. Being able to hook directly into the `deserialize` lifecycle allows me to intercept the data before it is set to the store, and ensure that the data's props are not being set directly on the store, but being wrapped in the models in my store.  Just using `bootstrap` this data has already been set on my store, and now I have to do more work to clean up my store and get it into the state I want.  See the [example](https://github.com/jdlehman/alt/blob/9430dcef13220c0b6e051a085eeba15231ed3bcc/docs/serialization.md#example), which is in the docs!

I think the other big area that `serialize` and `deserialize` are going to help is when users get data from the an external server and need to get format the JSON response into the structure that the store expects.  As a real world example: we have to solve this problem because while our app is new, the data model that backs it is over ten years old so we have to be able to receive data and tweak it into the format that the store expects (`deserialize`).  Then we can utilize snapshots as the data to send to an endpoint on our server to persist our data in the DB.  The endpoint is also legacy so we need to reformat the snapshot into the condition it expects too (via `serialize`).

My use case aside, this is just an example of what can be done.  In actuality the solution is so simple, yet gives so much direct power into the snapshot/bootstrap process that it is flexible enough to help a wide variety of solutions that individuals or companies may have based on existing constraints or desires for the app to behave in a certain way.

TLDR; `serialize` and `deserialize` hook in directly to where the data is being saved to the snapshot, or set to the store's state from the snapshot/bootstrap data, which gives users direct control and flexibility of how snapshots work within alt.  I think that the amount of control this provides for such a little amount of code (most of what I committed is docs/test changes), is a big win for alt.

goatslacker:
Lets break this down because I have a similar use case and issue.

For serialize: I'm currently using the `snapshot` lifecycle method.

```js
// immutable js example...
this.foo = Immutable.Map({a:1, b:2, c:3}); 

this.on('snapshot', () => {
  this.foo = this.foo.toJS();
});

this.on('bootstrap', () => {
  this.foo = this.foo.fromJS();
});
```

I also have another case where I'm using [LokiJS](https://github.com/techfort/LokiJS) to have an in-memory database. On snapshot I serialize the database so it gets saved properly and on bootstrap and re-initialize the database.

If the problem is that you'd like for a shortcut to set state (rather than manually set the state yourself) then this can probably be accommodated with the current API. If snapshot returns an object we set it, otherwise we do nothing. Although...

Letting you customize what to your snapshot should look like feels disingenuous because your store's actual state is not purely represented in the snapshot (only a facade of it) and this can lead to bugs later in the future when inspecting snapshots. This seems like another magic method.

Another case against having both `snapshot` and `serialize` available can get a bit confusing because you can then get yourself in a situation where you're accidentally using both snapshot and serialize and this can lead to hard to track bugs.

goatslacker:
As for deserialize.

Right now you can hook into both `init` and `bootstrap` and have it handle all of your needs. But the problem is that you have two functions that you need to hook into, and that the `init` function also gets called when the store initializes.

I'm open to solutions here. Perhaps we can introduce the deserialize method which will get called when the application state is being set before any other methods get called.

jdlehman:
I agree with you, probably the most negative thing that this can introduce is that it could make the lifecycle process more confusing.  We can aid this with a visual that shows exactly when the lifecycle methods get called, which would definitely help, but I think we might be able to work together to find the best solution.

The current lifecycle methods are great because they give you hooks into the process, but there is still some magic that alt does that you do not have control over, such as using `getState` as the source of what is stored in a snapshot, which is a good default, but I think there should be an escape hook to handle different use cases. Essentially you can use the current hooks to influence the process, but you don't have any direct control of the data that is saved in a snapshot, or back to the store.

The problem with the current `snapshot` solution you proposed is that you must alter the store itself to affect what data gets set in the snapshot.  It would be great if we could influence this without having to mutate the store to get the result we want.

I want to propose this change (based on some of your thoughts) to what currently is in my branch. We remove the `serialize` method. You are right, it is confusing to have both, that was one of my first thoughts, but I figured I would push everything up to get feedback before trying alternatives.  Instead, we can just use the `snapshot` method, but allow for an optional return.  If nothing is returned, we use the default `getState` to set the data in the snapshot.  This would look like the following: 

```js
const snapshot = (instance) => {
  return JSON.stringify(
    Object.keys(instance.stores).reduce((obj, key) => {
      const store = instance.stores[key]
      let customSnapshot
      if (store[LIFECYCLE].snapshot) {
        customSnapshot = store[LIFECYCLE].snapshot()
      }

      if (customSnapshot) {
        obj[key] = customSnapshot
      }
      else {
        obj[key] = store.getState()
      }
      // or if you don't mind the ternary operator
      // obj[key] = customSnapshot ? customSnapshot : store.getState()
      return obj
    }, {})
  )
}
```

Okay, so now we solve the same problem, but don't need to introduce another lifecycle method, `serialize`.  Now onto `deserialize`.

I still think that `deserialize` provides a nice hook that `bootstrap` does not accomplish, and I think the differentiation is less confusing than the difference between `snapshot` and `serialize`.  So I propose we add `deserialize` and modify `snapshot` to allow the flexibility of these escape hatches, without adding bulk or confusion to alt!

jdlehman:
Just rebased to remove merge conflicts.

jdlehman:
@goatslacker any more thoughts on this since I took your recommendations and changed the code?  Right now this just adds the `deserialize` hook and allows the `snapshot` to optionally return the data explicitly to be used in the snapshot.

Minimal code, but gives a lot of power/flexibility in these hooks IMO.

src/alt.js
```diff
@@ -219,18 +219,30 @@ const StoreMixin = {
...
     Object.keys(instance.stores).reduce((obj, key) => {
-      if (instance.stores[key][LIFECYCLE].snapshot) {
-        instance.stores[key][LIFECYCLE].snapshot()
+      const store = instance.stores[key]
+      let customSnapshot
+      if (store[LIFECYCLE].snapshot) {
+        customSnapshot = store[LIFECYCLE].snapshot()
       }
-      obj[key] = instance.stores[key].getState()
+      obj[key] = customSnapshot ? customSnapshot : store.getState()
```
goatslacker: this line can change to

```js
obj[key] = store[LIFECYCLE].snapshot ? store[LIFECYCLE].snapshot() : store.getState()
```

src/alt.js
```diff
@@ -219,18 +219,30 @@ const StoreMixin = {
 const setAppState = (instance, data, onStore) => {
   const obj = JSON.parse(data)
   Object.keys(obj).forEach((key) => {
-    assign(instance.stores[key][STATE_CONTAINER], obj[key])
-    onStore(instance.stores[key])
+    const store = instance.stores[key]
+    if (store[LIFECYCLE].deserialize) {
+      obj[key] = store[LIFECYCLE].deserialize(obj[key])
+      /* istanbul ignore next */
+      if(!obj[key]) {
```
goatslacker: rather than throw an error how about we assign if it returns otherwise we don't assign? this allows you to do anything you want in deserialize without having the pressure of having to set the state yourself.

docs/serialization.md
```diff
@@ -0,0 +1,86 @@
...
+permalink: /docs/serialization/
+---
+
+# Serialization
+
+The [snapshot](lifecycleListeners.md#snapshot) and [deserialize](lifecycleListeners.md#deserialize) lifecycle listener methods can be utilized separately or together to transform the store data being saved in a snapshot or the snapshot/bootstrap data being set to a store. Though they do not have to be used together, you can picture the parity between the two methods. You can transform the shape of your store data created in the snapshot with `snapshot`, which might be to get your data in a format ready to be consumed by other services and use `deserialize` to transform this data back into the format that your store recognizes.
+
+In the example below, we will show how this technique can be used to `serialize` complex model data being used by the store into a simpler structure for the store's snapshot, and repopulate our store models with the simple structure from the bootstrap/snapshot data with `deserialize`.
+
+## Snapshot
```
goatslacker: Ok we can rename this back to `serialize`

src/alt.js
jdlehman: Just to clarify, do you want to:

1. not assign anything to the store if there is no return
2. assign the default data, `obj[key]`, if there is no return

jdlehman: thanks for reviewing this btw! :smiley:

goatslacker: I'm thinking assign the default data if there's no return. It's the current default behavior so it's the least surprising of the two. Yes?

jdlehman: Agreed, I will make those changes, they all sound good to me.

src/alt.js
jdlehman: Didn't make this change. Because then it makes `snapshot` expect a return value.  Keeping it the way it is, is a bit more verbose, but always uses the default unless there is a valid return value.

I didn't catch it, but the tests did lol

goatslacker: Ah yes good point.

goatslacker: In that case how about const customSnapshot = ternary so that we don't mutate.

jdlehman: I like it.

jdlehman:
Not sure why coveralls always fires twice...

goatslacker:
Awesome

PR #76, opened 2015-03-14
Allow creating stores from a POJO

goatslacker:
This will allow you to create a store using a standard object.

There are some benefits to this approach:

You can use ES6 modules: https://gist.github.com/goatslacker/5a353916af65b3ee9987

You can use the ES3 Module pattern: https://gist.github.com/goatslacker/0f709f4436ed4890945d

And you can use other languages. This approach feels more flexible and less magical.


jdlehman:
:+1:

troutowicz:
Looks good, your es3 gist uses a `statics` property though, should probably update that to use `publicMethods`.

PR #78, opened 2015-03-14
Always emit a change

goatslacker:
Should we always emit an event automatically on bootstrap, recycle, and flush?

### DIFF
--- a/dist/alt-browser-with-addons.js
+++ b/dist/alt-browser-with-addons.js
@@ -1451,6 +1451,7 @@ var Alt = (function () {
           if (store[LIFECYCLE].rollback) {
             store[LIFECYCLE].rollback();
           }
+          store.emitChange();
         });
       }
     },
@@ -1483,6 +1484,7 @@ var Alt = (function () {
           if (store[LIFECYCLE].bootstrap) {
             store[LIFECYCLE].bootstrap();
           }
+          store.emitChange();
         });
       }
     },
--- a/dist/alt-browser.js
+++ b/dist/alt-browser.js
@@ -1195,6 +1195,7 @@ var Alt = (function () {
           if (store[LIFECYCLE].rollback) {
             store[LIFECYCLE].rollback();
           }
+          store.emitChange();
         });
       }
     },
@@ -1227,6 +1228,7 @@ var Alt = (function () {
           if (store[LIFECYCLE].bootstrap) {
             store[LIFECYCLE].bootstrap();
           }
+          store.emitChange();
         });
       }
     },
--- a/dist/alt-with-runtime.js
+++ b/dist/alt-with-runtime.js
@@ -471,6 +471,7 @@ var Alt = (function () {
           if (store[LIFECYCLE].rollback) {
             store[LIFECYCLE].rollback();
           }
+          store.emitChange();
         });
       },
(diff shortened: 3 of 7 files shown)


jdlehman:
I think this is a good default. As long as nobody has a good reason why they would not want their app to update from the state change, I think this is fine.

goatslacker:
I think so too. I'm going to drop this into a couple of apps and see how it reacts. If everything goes well I'll merge it in.

jdlehman:
I confirmed that it works in my alt apps. Makes sense, as I was wondering the other day why this wasn't already the default.

goatslacker:
My only concern is a potential case where you're not expecting it to emit a change and it does. I can't think of such a situation other than if you're listening on `componentWillMount` and attempting to do server-side rendering. Or if you've got store listeners setup with side effects on change.

jdlehman:
I still like this as the default, but we probably should provide a way to prevent the `emitChange` if not desired.  Here are some ideas off the top of my head:

1. have a property like `altSettings` on the store, that the user can set properties on (`altSettings` would be an object).  *The good:* This would make it easy to add properties in the future, and determine if the user is overriding defaults.  *The bad:* We would have a lot of gross if statements littered throughout the alt source.
2. We could allow the user to pass in an optional config object to `alt.createStore`. *The good:* More explicit than option 1, and all the other good things from option 1. *The bad:* Same as option 1.
3. We always `emitChange` unless the user defines a lifecycle hook method, then it falls on them to `emitChange`. *The good:* No magic. No gross if statements everywhere. *The bad:* If a user wants to use the defaults except they don't want to `emitChange`, then they have to add an empty lifecycle hook (that would be confusing and not ideal).

goatslacker:
How about we do what stores do? `return false` suppresses the event? (I haven't thought this one through)

jdlehman:
That was my first thought, except that `serialize` and `deserialize` allow returns.  We would have to return an object for those ones `{emitChange: false, data: {...}}`.

I think I like the store config the best out of what has been discussed (option 2 from my list).  I probably exaggerated a bit with "We would have a lot of gross if statements littered throughout the alt source."  In this specific case we would just have one per lifecycle method, but having a config option does open the door to flexibility, but at the price of complication, which we should be wary of. 

We could also be overthinking this. In this particular case I am hard pressed to think of an example that I would not want the change event to happen.

goatslacker:
Yeah with deserialize/serialize I'd think you would want it to emit a change regardless though. So it's safe to assume if you return something then it'll emit, otherwise if you return false it won't.

jdlehman:
Yep, was just about to say that after thinking it through.  If it sounds good to you, I say go ahead and always emit the change when false is not returned.  `deserialize`/`serialize` are a bit different in nature so it makes sense for them to behave differently.

goatslacker:
I'll amend the changes later tonight.

jdlehman:
I love how I write long comments, and right after I hit submit the answer hits me or I realize something incorrect with my logic. :frowning:

goatslacker:
It's not the first time I've thought about store options. I once had an implementation to disable bootstrapping at the app level and had that through an alt option. I ultimately decided against it because of the if statements that were starting to spawn everywhere.

PR #79, opened 2015-03-16
Fix exception if bootstrap non-existent store

bkniffler:
Prevent exception if trying to bootstrap with non-existent store.

### DIFF
--- a/src/alt.js
+++ b/src/alt.js
@@ -226,7 +226,7 @@ const StoreMixinEssentials = {
 const setAppState = (instance, data, onStore) => {
   const obj = JSON.parse(data)
   Object.keys(obj).forEach((key) => {
-    if(instance.stores[key]){
+    if (instance.stores[key]) {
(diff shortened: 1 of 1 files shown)


src/alt.js
```diff
@@ -226,8 +226,10 @@ const StoreMixinEssentials = {
 const setAppState = (instance, data, onStore) => {
   const obj = JSON.parse(data)
   Object.keys(obj).forEach((key) => {
-    assign(instance.stores[key][STATE_CONTAINER], obj[key])
-    onStore(instance.stores[key])
+    if(instance.stores[key]){
```
goatslacker: nit: whitespace

`if (instance.stores[key]) {`

bkniffler: Done!

jdlehman:
@bkniffler do you want to rebase this with master so we can merge it in? I can also rebase it if you prefer.

goatslacker:
b458450e9c46fe38a44f3a23a6dc0c1e08bbf012

PR #80, opened 2015-03-17
Update babel to latest version

jdlehman:
- If a user npm installed from prior package version they would still get this version of babel, which has some changes in compilation. This update ensures that everyone is on the current version to avoid confusion in `dist/` changes.


jdlehman:
I didn't realize this was an issue until I pulled down alt and did an`npm install` only to find that I already had `dist/` changes despite it being a fresh clone.

goatslacker:
:+1:

Issue #54, opened 2015-03-07
Snapshots of only certain stores


goatslacker:
Right now takeSnapshot takes the entire application snapshot. It would be nice if you could take a snapshot of a subset of your application.

PR #82, opened 2015-03-17
Allow `takeSnapshot` to specify stores

jdlehman:
- This enables snapshotting a subset of the app data
- `takeSnapshot` returns stringified JSON containing just the subset of data specified. The new snapshot is merged with the `LAST_SNAPSHOT` data to ensure that the last snapshot of the stores not specified in the subset is not lost. This also means that rollbacks will just work.
- Resolves Issue #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ var locationActions = alt.generateActions('updateLocation', 'updateCity', 'updat
 
 Stores are where you keep a part of your application's state.
 
-`alt.createStore :: Class, String -> Store`
+`alt.createStore :: Class, string -> Store`
 
 ```js
 class LocationStore {
@@ -529,7 +529,7 @@ Restart the loop by making your views kick off new actions.
 
 ### Snapshots
 
-`takeSnapshot :: ?...String -> String`
+`takeSnapshot :: ?...string -> string`
 
 Snapshots are a core component of alt. The idea is that at any given point in time you can `takeSnapshot` and have your entire application's state
 serialized for persistence, transferring, logging, or debugging.
@@ -538,7 +538,7 @@ Taking a snapshot is as easy as calling `alt.takeSnapshot()`. It can also take a
 
 ### Bootstrapping
 
-`bootstrap :: String -> undefined`
+`bootstrap :: string -> undefined`
 
 Bootstrapping can be done as many times as you wish, but it is common to use when initializing your application. The `alt.bootstrap()` function takes in a snapshot (JSON string)
 you've saved and reloads all the state with that snapshot, no events will be emitted to your components during this process, so again, it's best to do this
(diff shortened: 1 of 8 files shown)


src/alt.js
```diff
@@ -450,9 +451,14 @@ class Alt {
     }, exportObj)
   }
 
-  takeSnapshot() {
-    const state = snapshot(this)
-    this[LAST_SNAPSHOT] = state
+  takeSnapshot(...storeNames) {
+    const state = snapshot(this, ...storeNames)
+    if(this[LAST_SNAPSHOT]) {
```
goatslacker: Why assign when it's a subset? I think we should always set so it's predictable.

docs/takeSnapshot.md
```diff
@@ -7,13 +7,14 @@ permalink: /docs/takeSnapshot/
...
 
-> (): string
+> (...storeNames: ?string): string
 
-Take snapshot provides you with the entire application's state serialized to JSON.
+Take snapshot provides you with the entire application's state serialized to JSON, by default, but you may also pass in store names to take a snapshot of a subset of the application's state.
 
 Snapshots are a core component of alt. The idea is that at any given point in time you can `takeSnapshot` and have your entire application's state
-serialized for persistence, transfering, logging, or debugging.
+serialized for persistence, transferring, logging, or debugging.
```
goatslacker: haha, thanks

README.md
```diff
@@ -529,12 +529,12 @@ Restart the loop by making your views kick off new actions.
 
 ### Snapshots
 
-`takeSnapshot :: String`
+`takeSnapshot :: ?...String -> String`
```
goatslacker: this should be lowercase actually

src/alt.js
jdlehman: My reasoning for this was because of the following scenario.

1. I take a snapshot of my app, containing 5 stores or so.
2. Some change occur in my stores.  If I rolled back from here I would be back at my first snapshot. So far so good.
3. I take a partial snapshot of only 2 of my stores. 
4. More changes occur in my stores. If I rollback from here, the 2 stores I snapshotted will rollback, but my other stores will remain untouched.  This could be confusing behavior.  If store 5 (a store not in my second snapshot) had an error, so I try to rollback from there since I had taken a snapshot of it at some point, it would not recover back to that snapshot.  I am using assign to merge snapshots such that any stores that were in a previous snapshot, but not included in the new snapshot will not have their rollback data lost.

I understand what you are saying and would like it to always be predictable.  The problem is you could lose other store's rollback data.  This might be better solved in another ticket to make rollbacks more robust.  Curious to hear your thoughts on this scenario though.

goatslacker: Makes sense. In this case there's some small formatting nits.

src/alt.js
```diff
@@ -450,9 +451,14 @@ class Alt {
...
+      assign(this[LAST_SNAPSHOT], state)
+    }
```
goatslacker: } else {

src/alt.js
goatslacker: whitespace on the if stmt.

goatslacker:
:+1:

src/alt.js
jdlehman: I'll rebase and run the working linter on it to clean that stuff up

PR #84, opened 2015-03-18
Update eslint and fix linting errors

jdlehman:
- Update to latest version of eslint with much better es6 support. We no longer will need `babel-eslint`
- Fix all linting errors

### DIFF
--- a/dist/alt-browser-with-addons.js
+++ b/dist/alt-browser-with-addons.js
@@ -1197,7 +1197,9 @@ var StoreMixinListeners = {
         _this8.bindAction(symbol, listener);
       }
     });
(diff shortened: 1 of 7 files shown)


jdlehman:
Curious what people's thoughts are on running linting with Travis or if it should be a less formal process.

goatslacker:
Oh wow nice, I'd rather turn on eslint for travis if we can this way it catches style issues. We can actually use something like [ghooks](https://www.npmjs.com/package/ghooks) to [block pre-commit pushes](https://github.com/jshint/fixmyjs/blob/master/package.json#L58)

PR #89, opened 2015-03-19
Add store instance properties

goatslacker:
This opens the door for a couple of things:

Having direct access to the store model constructor makes for nice and easy testing (no more double exports) and it also makes for better introspection.

Having the names of the subscribed listeners is good for introspection.

These are properties rather than functions because I see no reason to hide them and mutating them shouldn't have terrible consequences.


troutowicz:
:beers: here's to no more double exports

jdlehman:
Amen to that.  We will have to update the test docs once this goes out

goatslacker:
We can have both as an option. I like the double exports if you're on ES6.  :) but I'll make sure to add this in.

goatslacker:
Updated with tests, this is good to go.

jdlehman:
Looks good!

PR #90, opened 2015-03-19
Fix subsequent snapshots

jdlehman:
- first snapshot works, but any snapshots after that are not correctly set
- `snapshot` returns stringified json, so in order to use assign, we must first parse the stringified json so that the state merges correctly
- add a test so this issue is checked in the future

### DIFF
--- a/src/alt.js
+++ b/src/alt.js
@@ -460,7 +460,7 @@ class Alt {
   takeSnapshot(...storeNames) {
     const state = snapshot(this, ...storeNames)
     if (this[LAST_SNAPSHOT]) {
-      assign(this[LAST_SNAPSHOT], state)
+      this[LAST_SNAPSHOT] = JSON.stringify(assign(JSON.parse(this[LAST_SNAPSHOT]), JSON.parse(state)))
     } else {
       this[LAST_SNAPSHOT] = state
     }
--- a/src/alt.js
+++ b/src/alt.js
@@ -241,14 +241,12 @@ const setAppState = (instance, data, onStore) => {
 
 const snapshot = (instance, ...storeNames) => {
   const stores = storeNames.length ? storeNames : Object.keys(instance.stores)
-  return JSON.stringify(
-    stores.reduce((obj, key) => {
-      const store = instance.stores[key]
-      const customSnapshot = store[LIFECYCLE].serialize && store[LIFECYCLE].serialize()
-      obj[key] = customSnapshot ? customSnapshot : store.getState()
-      return obj
-    }, {})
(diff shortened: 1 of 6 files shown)


src/alt.js
```diff
@@ -460,7 +460,7 @@ class Alt {
   takeSnapshot(...storeNames) {
     const state = snapshot(this, ...storeNames)
     if (this[LAST_SNAPSHOT]) {
-      assign(this[LAST_SNAPSHOT], state)
+      this[LAST_SNAPSHOT] = JSON.stringify(assign(JSON.parse(this[LAST_SNAPSHOT]), JSON.parse(state)))
```
jdlehman: This line is a bit gross. We could break it up into multiple lines to help make it more readable. We could also have the snapshot method return an object, and stringify that result later as it is only used in the `takeSnapshot` and `flush` methods. Or we could keep the line as is because one liners are nice in a way.

Thoughts, @goatslacker?

goatslacker: I like the second one, return obj from snapshot and stringify at flush and takesnapshot. You'll still have to JSON parse the last snapshot though.

goatslacker:
:+1: test is a great idea

src/alt.js
jdlehman: Sounds good to me, I will make the change in a bit. That was the one I liked as well

jdlehman: Okay, so that did not clean up as nicely as I was hoping.  My first attempt I ended up with this:

```js
  takeSnapshot(...storeNames) {
    const state = snapshot(this, ...storeNames)
    if (this[LAST_SNAPSHOT]) {
      this[LAST_SNAPSHOT] = JSON.stringify(assign(JSON.parse(this[LAST_SNAPSHOT]), state))
    } else {
      this[LAST_SNAPSHOT] = JSON.stringify(state)
    }
    return JSON.stringify(state)
  }
```

which I refactored into:

```js
  takeSnapshot(...storeNames) {
    const state = snapshot(this, ...storeNames)
    this[LAST_SNAPSHOT] = JSON.stringify(
      assign(JSON.parse(this[LAST_SNAPSHOT] || '{}'), state)
    )
    return JSON.stringify(state)
  }
```

jdlehman: Wondering if there is a reason we set `LAST_SNAPSHOT` to null initially, rather than just making it an empty object like we do with `INIT_SNAPSHOT`.  Doing so could clean this up even more.

goatslacker: setting it to INIT_SNAPSHOT is reasonable.

PR #94, opened 2015-03-27
use envify in script builds, to work in windows

mikekidder:
https://github.com/hughsk/envify#cli-usage

This makes the build scripts work in windows, no errors.


goatslacker:
Amazing. Thank you very much!

PR #97, opened 2015-03-27
Immutable js

goatslacker:
Just some thoughts and ideas on how to support immutable-js first class in alt with very little code.


ziflex:
What if detect type of store's state and depending that information return it to the caller ?
In this case there won't be need to add Immutable.js to the project and give users choice use it or not.
I would really like only one method for getting state `getState` to avoid extra complexity.

goatslacker:
@ziflex my thoughts on adding the `getImmutableState` was that it complements `setImmutableState`

src/alt.js
```diff
@@ -34,6 +34,13 @@ function uid(container, name) {
   return key
 }
 
+/* for now just detect whether it's instance of Immutable.js */
+function isImmutable(state) {
+  // let's check with that for now
+  // https://github.com/facebook/immutable-js/issues/421
+  return typeof state.toJS === 'function'
```
jdlehman: In theory I like this, but I am wondering if we want to bake this into alt, as it relies on the api of a specific library, one that is not even an alt dependency.

It also seems weird to me to support just one library when there are so many immutable libraries popping up these days.

ziflex: Well, we can make a dirty check - if object contains at least one functions - it will be detected as immutable data structure.

goatslacker: FWIW I don't mind supporting immutable-js as first class into alt, it has a lot of traction and there's been enough people asking for it. The trick is to do it in a minimal and elegant way.

goatslacker: If we end up having a util I'm thinking we can pass a special symbol which denotes if the store is immutable or not.

jdlehman: @goatslacker makes sense to me. I just always like to play devil's advocate to adding things to make sure we thing about them :smiling_imp:

ziflex: @goatslacker what will happen if during runtime store gets mutable data?

dist/alt-browser-with-addons.js
```diff
@@ -1317,15 +1329,24 @@ var createStoreFromObject = function (alt, StoreModel, key, saveStore) {
     dispatcher: alt.dispatcher,
     getInstance: function getInstance() {
       return storeInstance;
-    },
-    setState: function setState() {
+    }
+  }, StoreMixinListeners, StoreMixinEssentials, StoreModel);
+
+  if (StoreModel[IMMUTABLE] === true) {
+    StoreProto.setState = function (key, value) {
```
goatslacker: This can actually be DRYed up.

utils/ImmutableUtil.js
```diff
@@ -0,0 +1,71 @@
...
+
+  return store
+}
+
+function makeImmutableClass(StoreClass) {
+  function ImmutableClass() {
+    var _this = this
+
+    _get(Object.getPrototypeOf(ImmutableClass.prototype), 'constructor', this).call(this)
+    this.state = Immutable.Map({})
```
goatslacker: I need to take the things out of the instance variables here

jdlehman:
:+1: Looks good after all the cleanup you did!  I like this approach because it does not add anything super opinionated and I think adding this functionality in the util is the correct way to go.

ziflex:
Looks good, so for creating immutable stores we will have to use ImmutableUtil?

goatslacker:
I mean, you don't have to. You can use immutable-js right now without any of this, this just facilitates the setting up of the lifecycle methods and adds some sugar for `setState`.

There's a problem with this specific implementation though, `setState` only allows you to use `set` from Immutable, not the other useful methods like `merge`, `delete`, etc...

goatslacker:
Ok fixed, this needs docs.

ziflex:
I updated ReactStateMagicMixin like this in order to update immutable state properly:
```javascript
export default lodash.mixin(ReactStateMagicMixin, {
  altSetState: function() {
    this.replaceState(this.getStateFromStores());
  }
});
````

Maybe it needs to add this 'updated' mixin in Alt core ?

goatslacker:
Interesting, I thought `replaceState` was going away in React?

ziflex:
@goatslacker https://github.com/facebook/react/blob/ab7084f71d904d08543cfae2219630d331a2a0af/docs/docs/ref-02-component-api.md#replacestate

goatslacker:
> It may be removed entirely in a future version of React.

yeah :confused:

goatslacker:
Any last thoughts on API?

ziflex:
hmm It's confusing, that we use 'setState' but 'getImmutableState'... :(
ImmutableUtil.createStore is ok, but why we can not use regular 'getState' ?
Btw, maybe we we could just provide additional parameter to 'alt.createStore' which would specify whether the store is immutable. What do you think guys ?

jdlehman:
We might want to specify and provide a link in the docs that the util is using Facebook's Immutable.js library (so it is less confusing).  Another with the same API would work as well, so that should probably be mentioned too

jdlehman:
Another that still feels wrong that we are making changes in alt core.  I think it would be getter if the setImmutableState stuff could live in the util as well (have not looked deep enough to see if this is even possible, given the current implementation of set state).  So if this isn't easily doable, then I think it is fine.

troutowicz:
If it isn't possible, then perhaps we should look into making alt more pluggable :)

goatslacker:
- [ ] reduce amount of code in core alt.
- [ ] free up instance variables in immutable stores?
- [ ] remove any getImmutableState and setImmutableState mess

goatslacker:
And closing this one off too, I've got a better one coming.

PR #99, opened 2015-03-27
Move eslint config from package.json to .eslintrc

jdlehman:


jdlehman:
Any concerns cleaning up the `package.json` file with this move?

goatslacker:
I put it in package.json initially because I like having all the config in one place and not having an extra file in the repo, there's lots of files already in the root.

I do hate the formatting though.

...Musing about moving things to other repos...

Maybe we should move this eslint config to a separate repo, and npm install --save-dev it and then use it by referencing that. :smile_cat:

jdlehman:
Move all the things!

jdlehman:
It's all about modularity you know ;)

goatslacker:
haha. In this case though I'm semi-serious we can piggy-back off of an existing eslintrc that I'm sure exists somewhere and then we can override those settings inside our package.json.

https://github.com/airbnb/javascript will soon be adding a proposal for ES6-support maybe I should add an eslintrc and jscs file there.

jdlehman:
Oh, I'd be fine using that since we are already abiding by the airbnb style guide. Feel free to close this if that's the direction we go. The ultimate goal is to clean up the package.json file so it is easier to glance at, and either of these things will accomplish that.

goatslacker:
I think a more pressing need is to move the authors bit to the top, add your name, and move the license to the top as well.

jdlehman:
:+1: I'll close this and make a commit that moves the authors and license

PR #101, opened 2015-03-27
Add getImmutState method to stores

jdlehman:
- Uses immutable.js Record to return real immutable data (getState is
  not immutable for nested objects)
- Do not merge!  This is mostly to get a discussion around introducing Immutable.js directly to alt.


jdlehman:
In reference to Issue #56

jdlehman:
The problem with this approach is that we bring in a pretty large dependency.  We could leverage it in a lot of other places, which would be nice, but we need to decide if we want to go that route or not.

This could also just be a pattern that we put in the docs so people can easily integrate Immutable.js on their own.  I will put up another PR for a util later.

troutowicz:
All of these additions to alt are great, but unless its functionality directly related to alt I feel like the addition should be a util and kept out of alt's core.

ziflex:
@jdlehman but what your solution actually give us? You just wrap the plain JavaScript object with Immutable.Record completely loosing the real purpose of it. Why can I not do the same on my controller view level during fetching store's state?

From my perspective, you lost the idea of immutable data strucutre in Alt's stores, I think if we want use Immutable.js the only thing we need to do is - detect whether the store's state is instance of Immutable.js. If so, just return the reference of it. Otherwise keep original logic.

jdlehman:
@troutowicz yeah I think I agree with you. This is mostly just exploratory so we can get discussion on it.  I think I would prefer to make it the user's responsibility to bring in Immutable.js themselves.  Not really an alt concern.

@ziflex That is actually a great idea!

jdlehman:
Closing this in favor of PR #103. Much better to keep Immutable.js out of alt, but provide a hook to it, if it is used.  Great stuff @ziflex

PR #102, opened 2015-03-27
script update to run istanbul; windows compatibility

mikekidder:
Windows needs the full path for Istanbul to run _mocha.  Should be ok across platforms, may want to test.


goatslacker:
:+1: thanks!

PR #103, opened 2015-03-27
Added support for immutable data structure.

ziflex:


jdlehman:
I like this approach! :+1:

goatslacker:
I'm going to load this up and play with it.

goatslacker:
I'm turning off coveralls notifications.

goatslacker:
https://gist.github.com/goatslacker/1ddae9c27a822ccdfdd4

It should console.log:

```
This console log should fire
true
true
true
```

But the last condition is false:

```
This console log should fire
true
true
false
```

ziflex:
It's strange. Should be a way to fix it avoiding extra method.

ziflex:
I've just run the test using this example and it works.

ziflex:
With ```ImmutableUtil.makeImmutableStore``` Alt is losing it's beauty and simplicity :( And we'v got dependency.

goatslacker:
I think we need something like `ImmutableUtil` unfortunately because I'd like to decorate alt stores with Immutability. This decorator adds the lifecycle methods for you so alt.takeSnapshot and alt.bootstrap can work properly.

If I pull them into main alt then we'll have Immutable as a dependency which adds lots of bytes.

goatslacker:
Just ran the tests, I stand corrected on my point earlier. It's working fine :+1: I think I had an older version of Immutable, I haven't confirmed what the bug was on my end.

ziflex:
Super! Do we need more tests for it ?

goatslacker:
Yes, alt.takeSnapshot and alt.bootstrap should work.

src/alt.js
```diff
@@ -34,6 +34,13 @@ function uid(container, name) {
   return key
 }
 
+/* for now just detect whether it's instance of Immutable.js */
+function isImmutable(state) {
```
goatslacker: Another good thing about ImmutableUtil is that we can wrap the store in a way where we can make this check more robust.

jdlehman:
I always say it, but can we also add some docs for this?

I hate to see a great feature go unused because it isn't documented anywhere other than the source.

src/alt.js
ziflex: Yeah, I thought about moving this check out of the main source file.

goatslacker: The check being in here isn't what bothers me, it's the check for toJS being a function.

The con with this is that anyone can now define toJS as a function as part of their state and this will affect how getState() and setState behave.

test/immutable-store-state.js
```diff
@@ -0,0 +1,40 @@
...
+const alt = new Alt()
+
+const actions = alt.generateActions('fire')
+
+const assign = require('object-assign')
+
+const store = alt.createStore({
+  displayName: 'ImmutableStore',
+
+  state: Immutable.Map({}),
```
goatslacker: Since state is always a Map (am I wrong about this?) I was hoping to hide this behind ImmutableUtil as well.

src/alt.js
ziflex: Agree with you and I'm waiting for answer here:
https://github.com/facebook/immutable-js/issues/421

test/immutable-store-state.js
ziflex: Should it be initialized as Immutable from the beginning ? What if user will want use other immutable data structures from Immutable.js?

goatslacker: State is usually a key/value store like React's state so Map makes sense here. Also, maybe `Record`?

jdlehman: The thing with `Record`s is that they do not have any setters, it just throws an error, so I am not sure how useful it would be in a store, where you presumably would want to change your data on actions.

`Records` are nice because you can get the values from them without an explicit setter `obj.val`, but they are immutable.  So they are nice if you wrap mutable data in your store and return them as a `Record` such that they are immutable in the views, but I don't think they make sense in this context.

Maybe a `List`?  I think if we could make this configurable, it would be nice so we don't have to choose `Map` or `List`.  We could default to `Map` because that seems sensible, but provide an escape hatch to configure?

ziflex: @jdlehman Record has setter, but the main difference comparably with Map is that if you want to set value to key which wasn't defined in constructor, you'll get an exception.

Honestly, I don't think the it's good idea sticking to specific data structure.

jdlehman: Ah, I see that now in the docs. Thanks for setting me straight.  I still think it should be configurable despite whether or not we pick Record or Map as the default.

goatslacker:
I've copied over your work to #97

Issue #106, opened 2015-03-28
Provide your own dispatcher


goatslacker:
If you don't care for Facebook's Dispatcher you should be able to provide your own given that it conforms to the same interface `waitFor`, `dispatchToken`, `register`, `dispatch`.

This feature should be documented.

PR #108, opened 2015-03-28
Allow custom dispatcher to be specified

jdlehman:
- alt constructor takes config object that enables user to override defaults, in this case the dispatcher
- If dispatcher has interface with waitFor, dispatchToken, register, and dispatch functions, then it is valid and can be used, otherwise the default FB dispatcher is used as fallback
- This PR sets up a clean infrastructure to allow users to override other alt defaults
- Solves Issue #106

### DIFF
--- a/dist/alt-browser-with-addons.js
+++ b/dist/alt-browser-with-addons.js
@@ -1,6 +1,5 @@
 (function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.Alt = f()}})(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
 "use strict";
-
 /**
  * This mixin lets you setup your listeners. It is similar to Fluxible's mixin.
  *
@@ -80,7 +79,6 @@ module.exports = FluxyMixin;
 
 },{"./Subscribe":4}],2:[function(require,module,exports){
 "use strict";
-
 var Subscribe = require("./Subscribe");
 
 var ListenerMixin = {
@@ -111,7 +109,6 @@ module.exports = ListenerMixin;
 
 },{"./Subscribe":4}],3:[function(require,module,exports){
 "use strict";
-
 /**
  * This mixin automatically sets the state for you based on the key you provide
  *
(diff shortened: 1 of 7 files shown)


dist/alt-browser-with-addons.js
```diff
@@ -1353,11 +1349,24 @@ var createStoreFromObject = function (alt, StoreModel, key, saveStore) {
   return storeInstance;
 };
 
+var checkDispatcherInterface = function () {
+  var dispatcher = arguments[0] === undefined ? {} : arguments[0];
+
+  var requiredMethods = ["waitFor", "dispatchToken", "register", "dispatch"];
```
goatslacker: dispatchToken is a property and not a method on the Flux dispatcher

goatslacker: I'm actually not sure if we should even validate the dispatcher interface because there's no way to tell that register and dispatch will work correctly.

src/alt.js
```diff
@@ -332,9 +332,18 @@ const createStoreFromObject = (alt, StoreModel, key, saveStore) => {
...
+    return typeof dispatcher[prop] === 'function'
+  }, true)
+
+  return validInterface ? dispatcher : null
+}
+
 class Alt {
-  constructor() {
-    this.dispatcher = new Dispatcher()
+  constructor(config={}) {
```
goatslacker: :+1: I like this

dist/alt-browser-with-addons.js
jdlehman: That's a good point, I am going to remove this because it will be the user's responsibility to ensure they are using a dispatcher that will work.  No point in having extra code in alt to do that.

I will remove that, and add docs, then this should be good to go.

jdlehman:
Don't merge until I add docs

jdlehman:
Docs added, this is ready for another look.  Once this is in, I can easily add the ability to override JSON.stringify and JSON.parse with this new config object. :smile:

docs/altInstances.md
```diff
@@ -26,6 +26,23 @@ var flux = new MyAlt();
...
 # AltClass
 
+## AltClass#constructor
+
+> constructor(config: object) : Alt
+
+The alt constructor takes an optional configuration object. This is where you can configure your alt instance.
+
+```js
+var flux = new MyAlt({
```
goatslacker: s/MyAlt/Alt

docs/altInstances.md
```diff
@@ -26,6 +26,23 @@ var flux = new MyAlt();
...
+  dispatcher: new MyDispatcher()
+});
+```
+### Config Object
+
+The following properties can be defined on the config object:
+
+- *dispatcher*: By default alt uses Facebook's Flux [dispatcher](https://github.com/facebook/flux/blob/master/src/Dispatcher.js), but you can provide your own dispatcher implementation for alt to use. Your dispatcher must have a similar interface to the Facebook dispatcher including, `waitFor`, `dispatchToken`, `register`, and `dispatch`.
```
goatslacker: it would be nice to mention that an easy way to provide your own dispatcher would be to extend facebook's dispatcher.

```js
class MyDispatcher extends Dispatcher
```

and some example on extending that

goatslacker: My bad on the issue, `dispatchToken` is something that exists on the store not on the dispatcher itself. So we can remove this :)

src/alt.js
```diff
@@ -333,8 +333,8 @@ const createStoreFromObject = (alt, StoreModel, key, saveStore) => {
 }
 
 class Alt {
-  constructor() {
-    this.dispatcher = new Dispatcher()
+  constructor(config={}) {
```
goatslacker: small whitespace nit, space around the equal sign

goatslacker:
Some comments, but after that :+1:

Issue #55, opened 2015-03-07
Time Traveling


goatslacker:
Original:

With DispatcherRecorder you can replay certain events in your application. It would be cool if there was a util that would allow you to timetravel through your application but also allow you to specify if you'd only like to time travel certain stores.

Update:

I've got a util that lets you TimeTravel through your entire application's state. It uses alt (very meta) for the time travel bit. You have a set of actions: `revertTo`, `revertStoreTo`, and `destroyAll` which will revert to a specific point in time, revert a specific store to a point in time, or destroy all the stored events.

My only concerns at this point are with memory usage. This should probably use something like immutable or some cursors so we can keep memory usage pretty low.

I need help now fleshing out the rest of the API and most importantly if you've got feature requests then send them in :)

Open questions:

* If people want to be able to select certain values to be cursors then what does that look like? Which library should we use if any?
* What other features should this support?
* Thoughts on what the API should look like?

goatslacker:
Incoming PR for this

goatslacker:
cc @kenwheeler

goatslacker:
cc @gaearon who also might have opinions on this

kenwheeler:
+1 on immutable. Perhaps keep an array of actions and their payload values for replay? I'll think around this for a bit and report back.

kenwheeler:
What arguments would revertStoreTo take?

goatslacker:
`TimeTravelActions.revertStoreTo(name: string, id: string): undefined`

`name` is the store's name you'd like reverted although I could potentially use just the store itself here. `id` is a 7-character unique id that each snapshot will contain, this is used to identify to which state we should revert to.

```js
// if you defined a store like this
var myStore = alt.createStore(MyStore, 'MyStore')

time.getActions('TimeTravelActions').revertStoreTo('MyStore', 'abcd123')

// also possible:
time.getActions('TimeTravelActions').revertStoreTo(myStore, 'abcd123')
```

`TimeTravelActions.revertTo(id: string): undefined`

Reverts the entire application to the `id` provided.

`TimeTravelActions.destroyAll(): undefined`

Removes all snapshots from cache.

jareware:
This is really cool stuff. :) I'm not completely sure I understand why this would require any extra components, though, given how we already can `takeSnapshot()` after every dispatch. Shouldn't "time travel" be as easy as calling `bootstrap` with whatever snapshot you have previously taken?

ziflex:
[...]

Issue #53, opened 2015-03-07
Better rollback system


goatslacker:
Currently rollback only works if you've recently taken a snapshot. Perhaps there can be a way to enable automatic rollbacks in case something goes wrong. Ideally this would add minimal code.

jareware:
Neat idea! A concrete use case where this would be cool is an error during dispatch to stores A & B, but say A has been updated but B not, and it errors out. This could make store updates atomic, which would be pretty braggable.

PR #109, opened 2015-03-29
Add before and after hooks

goatslacker:


jdlehman:
This is great, I am already beginning to think of ways I can use this.  My only suggestion is to add an example.

I also think this opens the door for Issue #53.  Save state in the `before` hook with a snapshot, and then rollback in `after` if there was a problem.

goatslacker:
You read my mind :) This also opens the door for #55 since we can now save specific values and roll those back.

PR #112, opened 2015-03-29
Add note about exportPublicMethods in README

rocketraman:
In the README add the alternative approach of using exportPublicMethods in a store.


README.md
```diff
@@ -278,6 +278,8 @@ var locationStore = alt.createStore(LocationStore)
 locationStore.myPublicMethod()
 ```
 
+Another explicit, less magical, alternative to setting a method as static is to [declare](http://goatslacker.github.io/alt/docs/createStore/#storemodelexportpublicmethods) `exportPublicMethods` in the constructor.
```
goatslacker: Actually we can replace this whole section with `exportPublicMethods` and how to use it.

rocketraman:
> Actually we can replace this whole section with exportPublicMethods and how to use it.

Ok, let me take a stab at it.

rocketraman:
@goatslacker Rebased with recommended change.

Issue #52, opened 2015-03-07
Future of Alt


goatslacker:
This is just to openly discuss any ideas on where you'd like to take alt or what you'd like to do with it. Perhaps actionable tasks can come out of this discussion.

creatorrr:
With [React v0.13](http://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html) the use of ES6 classes will be available. Also, according to facebook/react#1380 mixins will not be supported for ES6 classes like they are today with createClass.

So, keeping those in mins,
 1. Are there any plans for supporting `alt` with ES6 classes once React v0.13 lands (which is coming soon)?
 2. Consideration should be made for alternatives to implement the functionality currently offered via mixins.

goatslacker:
You can already use alt with `React.Element` in v0.13. It seems like mixins are going to be left to be addressed at the language level, if there's a solution that pops up though alt will pick it up and support it. 

It was mentioned that there will be [first-class support for flux store subscriptions](http://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html#mixins), that's an interesting development.

bjyoungblood:
With actions that fetch data asynchronously, it seems to be a pretty common pattern to have associated success and failure actions. It might be worthwhile to explore a way to auto-generate those.

`this.generateActions` could possibly be modified to support an API [similar to Reflux's](https://github.com/spoike/refluxjs#asynchronous-actions).

To take this a bit further, I think it's common for an action to create a Promise that calls a corresponding success or failure action with its result. [MartyJS achieves this](http://martyjs.org/guides/action-creators/index.html) by creating the actions and triggering one of them for you with the result of the promise.

RnbWd:
Maybe use a mixin pattern like this? https://gist.github.com/sebmarkbage/ef0bf1f338a7182b6775

creatorrr:
@RnbWd Interesting but how would this work with multiple mixins overriding the same methods?

creatorrr:
@goatslacker Another question for alt is whether it's better to keep it to a small and extensible core or expand it to provide a decent feature set for developing frontend apps. IMHO, the direction of this conversation will depend on that.

RnbWd:
@creatorrr having multiple mixins overriding the same methods sounds like a bad idea.. honestly I don't know.  Do the current mixins override the same methods? I was just passing along a tweet I saw @gaearon post.

[...]

PR #115, opened 2015-03-30
Implements connectToStores()

cpsubrian:
First stab at the 'higher order component' discussed in #52. I'm open to feedback on any language or implementation details. I'm pretty new to React and especially new to testing React code, so let me know how everything looks

- Tests (with new ones) all pass
- Linter passes
- Coverage is **NOT at 100%**
  - There is some transpiled code that is not reached and I'm not sure of the best way to handle that. 
  - See screenshot below.

Note: I had to rework the `ReactComponent` helper a bit to facilitate my tests. I'm pretty sure I fixed a typo/error with the nesting of the static property copying inside the mixin copying.


cpsubrian:
Coverage:

screenshot 2015-03-30 13 09 39

utils/connectToStores.js
```diff
@@ -0,0 +1,94 @@
...
+    componentWillUnmount () {
+      stores.forEach(store => store.unlisten(this.onChange))
+    },
+
+    onChange () {
+      this.setState(Component.getPropsFromStores(this.props))
+    },
+
+    render () {
+      return React.createElement(Component, Object.assign({}, this.props, this.state))
```
cpsubrian: Noticed you're not using JSX in most places, but we could replace this with (what I use):

```jsx
render() {
  return <Component {...this.props} {...this.state} />
}
```

troutowicz: This is personal preference of course, but I tend to not use JSX in utilities, only in components. Makes it easier to understand the purpose of the file for me atleast.

cpsubrian: Yep, not big deal to me either way which is why I tried to stick to the common style.

cpsubrian: I'm gonna go out on a limb and guess that `Object.assign` is very similar to however spread attributes are implemented anyway :)

troutowicz: It's what I used when I modified your util for a personal project yesterday. :) It just returns a new object with the properties and state together.

cpsubrian:
Pushed a commit that should fix the lint error.

troutowicz:
What do you guys think about including bootstrap ability in this util.

```javascript
const alt = require('../alt');

...

componentWillMount() {
  if (this.props.altStores) {
    alt.bootstrap(JSON.stringify(this.props.altStores));

    // force state update this render cycle
    this.setState(Component.getStateFromStores(this.props));
  }
},

...
```

cpsubrian:
Looks like I need to read-up/try-out bootstrapping :)

troutowicz:
It's useful with an isomorphic app and passing initial data from server to clients.

utils/connectToStores.js
```diff
@@ -0,0 +1,94 @@
+'use strict'
+import React from 'react'
```
goatslacker: This is causing React to be included in the alt-browser bundle!!!! :astonished:

cpsubrian: Should we just have the user pass in React?

goatslacker: Probably assume people have react (window.React) in the browser version.

cpsubrian: 09b64ec removes React from the build, but not sure if `window.React` gets picked up. Need to work up a client-side test I guess.

cpsubrian:
ee2c04a makes some annoyingly trivial syntax changes such that the coverage is back at 100%. Guess we can't have all the ES6 without specific tests for the shims :)

troutowicz:
I'll post my modified version here in case it adds something worth while.

[connectToStores.js](https://github.com/troutowicz/geoshare/blob/master/app/utils/connectToStores.js)

cpsubrian:
@troutowicz Thanks! I intend to look into the bootstrapping with I get a chance.

Curious why you prefer the name `getStateFromStores` vs. `getPropsFromStores`? In the context of where the method is defined you are always 'updating' the props of the wrapped Component.

troutowicz:
We are always passing in data via props, but the data itself is coming from a store. Stores contain state. Either way would work though.

goatslacker:
I've rebased your commits into #177 lets get this in :dart:

PR #116, opened 2015-03-31
Fix react-hot-loader from reloading stores

### LINKED ISSUE #85: Changing stores with react-hot-loader
When making changes to a store don't create a new store.

goatslacker:
This fixes #85

### DIFF
--- a/src/alt.js
+++ b/src/alt.js
@@ -364,21 +364,23 @@ class Alt {
     let storeInstance
     let key = iden || StoreModel.name || StoreModel.displayName || ''
 
-    if (saveStore && (this.stores[key] || !key)) {
-      /* istanbul ignore else */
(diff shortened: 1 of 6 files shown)


src/alt.js
```diff
@@ -364,22 +364,24 @@ class Alt {
...
+          `A store named ${key} already exists, double check your store ` +
+          `names or pass in your own custom identifier for each store`
+        )
+      }
     }
 
+    key = this.stores[key] && StoreModel === this.stores[key].StoreModel
+      ? key
+      : uid(this.stores, key)
+
```
jdlehman: nit. Once a ternary has to be on multiple lines, does it make more sense to use an if statement?

goatslacker:
Doesn't really do what I thought it did. This needs more work

PR #127, opened 2015-04-02
fix AltContainer using ES6 concise function

wonderfulll:
Currently it throws 
```
AltContainer.js:107 Uncaught SyntaxError: Unexpected token (
```
when using AltContainer.

Another option would be to add 
```
  "browserify": {
    "transform": [
      "babelify"
    ]
  },
```
to package.json

### DIFF
--- a/components/AltContainer.js
+++ b/components/AltContainer.js
@@ -104,7 +104,7 @@ var AltContainer = React.createClass({
     }
   },
 
-  addSubscription(store) {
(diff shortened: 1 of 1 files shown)


goatslacker:
How silly of me. Thanks. I'll put out a patch tonight

goatslacker:
`npm 0.15.1`

Thanks again!

PR #130, opened 2015-04-03
Check context is not null

jellybob:
I'm not sure why, maybe its React 0.13 related, but the AltContainer context was null on my application starting, which caused it to blow up when trying to call `this.context.flux`. This change adds a null check there.

### DIFF
--- a/components/AltContainer.js
+++ b/components/AltContainer.js
@@ -116,8 +116,8 @@ var AltContainer = React.createClass({
 
   getProps: function () {
     return assign(
-      this.context.flux || this.props.flux
-        ? { flux: this.context.flux || this.props.flux }
+      (this.context && this.context.flux) || this.props.flux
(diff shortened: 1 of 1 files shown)


goatslacker:
Can you provide a failing test case for this?

jellybob:
If you can point me in the right direction to do so I can - I had a quick look, but couldn't find any tests for AltComponent in the `tests` directory.

goatslacker:
Sorry I thought I had answered this: https://github.com/goatslacker/alt/blob/master/test/store-listener-component-test.js the tests are located in that file. Feel free to create a new file though or just point me to a broken gist or share a broken code snippet.

goatslacker:
@jellybob does this fix the issue? 280b8fac252f9b9d2c035ec42cee4ec5b4c83bd5 if we check prop first?

goatslacker:
Closing in the interim.

PR #131, opened 2015-04-04
changed navbar from pure to bootstrap and made it responsive

jbhatab:


jdlehman:
Looks great! Thanks for working through this, it will help make the mobile experience much better on our site.

jdlehman:
Just deployed these changes so they are now live.

PR #132, opened 2015-04-05
Add immutable docs

jdlehman:
- Add documentation on examples of using Immutable.js with alt
- Create folder structure to add more docs on using other immutable libs

### DIFF
--- a/docs/immutable/immutablejs.md
+++ b/docs/immutable/immutablejs.md
@@ -123,3 +123,44 @@ class MyComponent extends Component {
   }
 }
 ```
+
+## Serializing/Deserializing Immutable Data
+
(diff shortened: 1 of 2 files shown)


jdlehman:
Would like a quick read over to make sure these make sense.

docs/immutableExamples/immutablejs.md
```diff
@@ -0,0 +1,125 @@
+# Using Immutable Data Structures
```
goatslacker: can we change the directory name to just `immutable`?

goatslacker:
We should add a note here on serializing/deserializing immutable data.

goatslacker:
Otherwise this is :+1:

docs/immutable/immutablejs.md
```diff
@@ -0,0 +1,166 @@
...
+// MyStore.js
+import {Map} from 'immutable';
+
+class MyStore {
+  constructor() {
+
+    this.on('serialize', () => {
+      return {
+        data: {
+          prop1: this.data.get('prop1'),
```
goatslacker: can't you use fromJS/toJS here or no?

jdlehman: Yep, that is actually much cleaner. I will update that. Thanks :)

PR #141, opened 2015-04-11
Allow passing in args to store and actions constructors

goatslacker:
Alt's first breaking change.

I'm thinking we should be able to pass in custom items into the store and actions constructors so this would make the following changes:

`alt.createStore` just takes a StoreModel and a name, all other arguments are passed to the store's constructor.

`alt.createActions` takes in the ActionsClass, an optional object to export actions to, all other args are passed to the action's constructor.

`alt.addActions` is name, class, and ...args

`alt.addStore` is name, class, and ...args

This removes the saveStore boolean that was previously present for addStore and createStore, if you wish to create a store that will not be part of bootstrapping then a new method `alt.createUnsavedStore` has been added.

The other breaking change is that actions and stores no longer have the alt instance passed in as the first argument implicitly, you now have to explicitly pass it in. This shouldn't be a huge deal since it's available through `this.alt` anyway.

This adds ~20LOC to core.

Thoughts?

### DIFF
--- a/dist/alt-browser-with-addons.js
+++ b/dist/alt-browser-with-addons.js
@@ -1072,7 +1072,7 @@ function NoopClass() {}
 var builtIns = Object.getOwnPropertyNames(NoopClass);
 var builtInProto = Object.getOwnPropertyNames(NoopClass.prototype);
 
-var getInternalMethods = function (obj, excluded) {
+function getInternalMethods(obj, excluded) {
   return Object.getOwnPropertyNames(obj).reduce(function (value, m) {
     if (excluded.indexOf(m) !== -1) {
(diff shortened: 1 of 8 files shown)


jdlehman:
:+1:  this is along the same lines of what we are doing by passing the config object to alt.  I really like the escape hooks for configurability, and it also makes it easier for us to add features/let users customize behavior in the future.

goatslacker:
Cool. 0.15.5 will be last patch and then I'll merge this in for 0.16.0

jareware:
:+1:, dependency injection through constructor arguments is both common and very simple, so this sounds great.

PR #142, opened 2015-04-11
Catch errors in stores

goatslacker:
This adds a try/catch to alt's dispatch for purposes of allowing you to catch errors in the store.

If your store class contains a `failedDispatch` method you'll receive the error and then you can deal with it.

This opens the door for better rollbacks/atomic transactions.

If the store does not contain a failedDispatch method then the error throws so it is not swallowed.

Adds ~11 LOC to alt core.

My only concern is that now every dispatch call is wrapped in a try/catch.

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -651,6 +651,28 @@ class Store {
 }
 ```
 
+`error` is called whenever an error occurs in your store during a dispatch. You can use this listener to catch errors and perform any cleanup tasks.
+
+```js
+class Store {
+  constructor() {
+    this.on('error', (err, actionName, payloadData, currentState) => {
+      if (actionName === MyActions.fire) {
+        logError(err, payloadData);
+      }
+    });
+
+    this.bindListeners({
+      handleFire: MyActions.fire
+    });
+  }
+
+  handleFire() {
+    throw new Error('Something is broken');
+  }
+}
+```
+
 ### Single Dispatcher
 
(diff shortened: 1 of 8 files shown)


jareware:
This is *exactly* what I had when I was still rolling my own flux-like architecture (before getting to know and jumping onboard alt). The only difference being that my `handleDispatchError()` was a global one, so that you could implement an `ErrorStore` with that method, from which the UI could retrieve any failures.

The use cases (which I admittedly never got to) were - for example - being able to implement form validation so that you could continuously try to dispatch the save action for that form, and in case it fails, roll back automatically and show validation errors. The grand idea of course being being able to reuse the logic/code for "doing X" and "seeing if I'd be able to do X if I tried".

Anyway, this seems like a very useful addition. Well-defined failure models aren't exactly the strong suit of most JS FW's, so :+1:.

jdlehman:
I like this change as well, as it provides a nice hook to deal with errors.

There are a couple of things I want to throw out there to consider:
- Can we pass the error into the args of the `afterEach` call instead of adding the `failedDispatch` method?  This would reduce code in the core, and makes sense as the the `afterEach` call is already being called after each dispatch.  However, it might be nice to have a method just for dealing with errors to keep that logic completely separate.
- If not should we move this to a lifecycle hook instead of a method directly on the store model?

goatslacker:
I like the separate method because then it isolates the error handling. 

Yes to lifecycle hook. I added this to the store by mistake and then realized that beforeEach and afterEach should be lifecycle as well.

PR #143, opened 2015-04-11
Atomic transactions

goatslacker:
This adds some initial atomic transactions util

It's pretty barebones in that it only supports POJOs. But it's a proof of concept.

This PR is rebased on top of: #142


goatslacker:
cc @jareware

goatslacker:
Fixes #53 somewhat

jareware:
So would failures during a single action in both stores `A` and `B` in fact trigger 2 subsequent rollbacks? Does this matter?

goatslacker:
Calling `rollback` repeatedly doesn't make a difference. Once it rolls back to the last snapshot it stays there until a new snapshot is created.

PR #159, opened 2015-04-14
Defer 'Store#emitChange' until promise resolve

mehcode:


goatslacker:
This is fairly clean. Some initial thoughts: this needs to check if Promise exists, this only works for one store?

src/alt.js
```diff
@@ -79,8 +79,12 @@ class AltStore {
       if (model[LISTENERS][payload.action]) {
         this[SET_STATE] = false
         const result = model[LISTENERS][payload.action](payload.data)
-        if (result !== false && this[SET_STATE] === false) {
-          this.emitChange()
+        if (this[SET_STATE] === false) {
+          if (result instanceof Promise) {
```
goatslacker: I've included this change in https://github.com/goatslacker/alt/pull/234

mehcode: Looks awesome; thanks for taking this and making it much better. :+1:

mull: I know this is an old commit that doesn't seem that it'll get merged it. Just an FYI that `object instanceof Promise` fails on Chrome for native promises.

goatslacker: :+1: we have an isPromise check.

PR #162, opened 2015-04-14
Make setState batch-able

### LINKED ISSUE #151: make setState aggregate calls
So we don't dispatch multiple calls lets just batch up all setState calls and emitChange once the dispatch is finished.

goatslacker:
Fixes #151

### DIFF
--- a/src/alt.js
+++ b/src/alt.js
@@ -15,7 +15,7 @@ const LAST_SNAPSHOT = Symbol('last snapshot storage')
 const LIFECYCLE = Symbol('store lifecycle listeners')
 const LISTENERS = Symbol('stores action listeners storage')
 const PUBLIC_METHODS = Symbol('store public method storage')
-const SET_STATE = Symbol()
+const STATE_CHANGED = Symbol()
 const STATE_CONTAINER = Symbol('the state container')
 
 const GlobalActionsNameRegistry = {}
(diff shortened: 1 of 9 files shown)


goatslacker:
Batches setState calls, allows for a function to be passed in. Adds ~23 LOC.

goatslacker:
cc @jdlehman thoughts?

src/alt.js
```diff
@@ -77,7 +99,8 @@ class AltStore {
         )
       }
       if (model[LISTENERS][payload.action]) {
-        this[SET_STATE] = false
+        this[IS_DISPATCHING] = true
```
jdlehman: do we really need this? can't we use `dispatcher.isDispatching()` to check if it is currently dispatching instead?

jdlehman:
I made one comment inline, everything else looks good to me.  I really like the option to pass in a function as well as an object.  It gives nice flexibility without adding too much extra code.

src/alt.js
goatslacker: Good question. Because someone may have registered themselves to the dispatcher or are using a lifecycle hook, in that case if you use setState it should still emit a change.

Example:

```js
function FooStore() {
  this.state = { calls: 0 }
  this.alt.dispatcher.register(function (payload) {
    this.setState({ calls: this.state.calls + 1 })
  });
}
```

jdlehman: thanks, makes sense to me

goatslacker: Although if its in a lifecycle hook or another dispatch call then it loses its batching benefits...hmmm maybe there is a better way.

PR #164, opened 2015-04-15
Before after lifecycle

goatslacker:


docs/lifecycleListeners.md
```diff
@@ -119,3 +119,15 @@ class Store {
   }
 }
 ```
+
+## beforeEach
+
+> (action: string, data: mixed, state: object): undefined
+
+This method gets called, if defined, before the payload hits the action. You can use this method to `waitFor` other stores, save previous state, or perform any bookeeping. The state passed in to `beforeEach` is the current state pre-action.
```
jdlehman: I might add a sentence clarifying that the data is the action's payload to help avoid any confusion (might even help to s/data/payload).

jdlehman:
:+1:

PR #166, opened 2015-04-16
Expose AltContainer in alt-browser-with-addons.js

danott:
I'm using a downstream version of `dist/alt-browser-with-addons.js`. I'm wanting to use `AltContainer`, but have no way of `require`'ing it, because I'm not in a node ecosystem. So my solution was to add `AltContainer` to `Alt.addons`.


danott:
I realized I could do everything I wanted with `ReactStateMagicMixin`, so this may or may not be a good idea.

goatslacker:
This is a good idea, but we have to exclude React from the build.

PR #167, opened 2015-04-16
Shuffle the browser addons

goatslacker:
Removes all the mixins and adds AltContainer + a few extras.


goatslacker:
@danott

danott:
@goatslacker wow, that was quick! What's the motivation for removing the mixins from `addons`? Seems like it could break things for downstream users? (Including my own project, and the solution I rolled with `ReactStateMagicMixin`)

goatslacker:
The motivation is to reduce bundle size. Generally you should favor the component over the mixins since the component does the same thing + more and mixins don't work if you're using class syntax with React 0.13.

It shouldn't break anything if I push this as 0.16.

I could add the mixins back in though, I don't have a very strong opinion here.

danott:
I tried using `dist/alt-browser-with-addons.js`, but it chokes in the browser on the `var React = require('react')`. I'm looking into the `browserify` docs to see if there is anyway to have that `require` return a variable defined at `window.React`

goatslacker:
Yeah just tried it and that needs fixing. I'll figure that out, should be simple

iam4x:
@danott https://github.com/thlorenz/browserify-shim about your err

danott:
@goatslacker just gave this another shot this morning, and everything works as expected using `AltContainer` instead of mixins. And it was a super easy change to make. Thanks for adding this into the addons!

goatslacker:
Cool I'll merge this in then.

PR #168, opened 2015-04-16
README: Added "function" to regular old JavaScript code

sahat:
Added missing `function` keyword, since in its current form it only works with an ES6 transpiler.


goatslacker:
Thanks!

PR #170, opened 2015-04-17
Refactor alt.js into multiple files

jdlehman:
This commit splits up the extremely large alt.js file into multiple files, grouped by functionality. This should help make the code base easier to read and make changes to in the future (and hopefully help make it less intimidating for future contributors)


src/store/createStore.js
```diff
@@ -0,0 +1,118 @@
+'use strict'
+
+import assign from 'object-assign'
+import AltStore from '../../src/store/AltStore'
```
jdlehman: I think we need to update some of the build scripts so that we can do `import AltStore from './store/AltStore'` etc.

Wasn't sure of the best way to do this with browserify (especially since the build files need to be consumable by the user)

goatslacker:
It's like we're connected somehow. I had been thinking about this the past few days and I had started something in a local branch but didn't get very far. Thanks :D

src/shared/symbols.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+import Symbol from 'es-symbol'
+
+export const ACTION_HANDLER = Symbol('action creator handler')
```
goatslacker: I hate to expose these because then you'd have a lot of access to alt's internals :confused:

jdlehman: That's a good point. Any ideas for providing access in multiple files?

Will these symbols be equivalent?
```js
// file1.js
const ACTION_HANDLER = Symbol('action creator handler')

// file2.js
const ACTION_HANLER = Symobl('action creator handler')
```

I assumed they will not be, but I think you raise a great point!

goatslacker: They won't. If we use `Symbol.for` they would be but then they're global symbols.

jdlehman: Hmm, I can't think of any other way to get around this since they are used by most parts of the app.

It might not actually be that bad of a thing because the built files in `dist/` won't give users direct access to these symbols anyway.  So the Symbols are exposed to the src only, but not to the end user.

goatslacker: but wouldn't you be able to require deep into the compiled source and pull these out?

goatslacker:
This is actually not working for me in a regular environment because src/ is not transpiled.

src/shared/symbols.js
jdlehman: Not to my knowledge.

The src files that gets built are `src/alt.js` or `src/alt-with-addons.js`.  Both of these may import numerous things, but all that is accessible is what they export for use, which is just Alt.

Essentially it is why we need this line in the addons file:
```js
Alt.addons = {
  ActionListeners,
  AltContainer,
  AltManager,
  DispatcherRecorder,
  atomicTransactions,
  chromeDebug,
  makeFinalStore,
  withAltContext
}
```
because importing the addons does not make them available so we have to attach them to the Alt object that we are exporting.

goatslacker: On second thought this isn't so bad. If people want to deep-require it then so be it.

jdlehman:
Added another commit that should handle the src being transpiled... Not sure if this is the optimal way to do this.

Only issue is now the coverage is off because it is looking at the transpiled output which has a lot of code from babel to handle error messages etc, which does not get hit by our tests (and should not be).

dist/alt.js
```diff
@@ -1,39 +1,1211 @@
+(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.Alt = f()}})(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
```
goatslacker: We shouldn't include a pre-built browserify file, webpack starts bitching about this.

I think the sensible thing to do would be to 

* transpile everything in src into lib
* change the main from dist/alt.js to lib/index.js or lib/alt.js or whatever
* remove dist/alt.js and only keep dist/alt-browser and with-addons which will be pre-browserified bundles

??

jdlehman: I like that approach :+1:

goatslacker: What about this for file structure:

```txt
src/
  utils/ (future home of all the utils)
  alt/
    index.js (the main file, maybe contains the class Alt?)
    createStore.js (createStoreForObject and Class)
    createActions (the actions stuff)
    utils/ (the helpers stuff goes here, the setAppState/snapshot stuff, the uid stuff, mixins for stores, etc...)
      ...
    symbols/
      symbols.js
```

goatslacker:
Just fixed all the merge conflicts. I'll stay out of here for a bit.

lib/alt/index.js
```diff
@@ -0,0 +1,366 @@
+"use strict";
```
goatslacker: I'm thinking we don't check these files in but rather just have them ran on pre-publish

jdlehman:
Thanks for cleaning this up @goatslacker :beers:

goatslacker:
That rebase was really ugly :imp:

goatslacker:
Thanks for getting this started, it was about time. Next is the tests, I've started it like 4 times already and have given up.

jdlehman:
Yep, I have a couple half started branches with the tests too.  It won't be easy, but it will be worth it. That index test is crazy

goatslacker:
I'm having second thoughts about this:

* It's tough to tell how big alt is at a quick glance. Before I used to be able to open up dist/alt.js and I would get an idea by LOC. This could be remedied by creating a quick gzip check of the build.

* More importantly, this makes the browser builds way, way larger. I'm not sure how much just yet in terms of bytes but in terms of LOC it's 500+ more.

I do like the structure and the way things build now so idk.

jdlehman:
There are definitely trade offs, but IMO I think that the readability is a huge win.  I think the easier we can make it for people to contribute/understand the codebase, the more we win in the long run.

goatslacker:
So the difference is 1115 LOC vs 661 LOC. :scream: 

Current minned: 23kb. Gzip: 6kb.
1File minned: 18kb. Gzip: 5.6kb. (not much savings)

Maybe there's some build magic we can do so we can get the best of both worlds.

Maybe we can keep it in separate files but concatenate them :trollface: reminds me of the old days. I'll keep thinking of solutions. Maybe I should just reduce the amount of files and that'll help.

frederickfogerty:
I assume you're externalising all the babel code so it's only included once? http://babeljs.io/docs/usage/external-helpers/

goatslacker:
We're not (only for testing) not sure how much this would help or hurt in this case I only want to include the babel-helpers we use.

frederickfogerty:
To be honest this is something that babel/webpack should be doing on its own, I'm surprised it doesn't.

Issue #158, opened 2015-04-13
Crazy question: Change state API?


goatslacker:
Do people like managing state through the instance properties or should we adopt a React 0.13-like API where state is set on a `state` property?

Now:

```js
class TodoStore {
  constructor() {
    this.todos = []
  }
}
```

React 0.13-like

```js
class TodoStore {
  constructor() {
    this.state = { todos: [] }
  }
}
```

I initially went with the instance properties approach because back in the old days of writing JavaScript where we'd have data stores we would create them as a function and constructor and then keep the state as instance properties. I felt that just changing an instance property was familiar enough to everyone and had less boilerplate.

I'm considering this new change because of a few things:

* Parity with React.

  Although alt prides itself on not being tied down to React, lets face it, Flux and React are like Peanut Butter and Jelly. Having the same API would make things smoother.

* Private instance variables.

  Currently you can have your private instance variables, but you'll have to use semi-private Symbols or wait for TC-39 to add language features to add private state to a class. By having this change anything not specified in `state` is private which can be nice.

* Stores are automatically serializable so with `state` you can define what is serialized right from the start.

  This is a big reason. Say for example you're using something like firebase.

```js
class TodoStore {
  constructor() {
    this.firebase = new Firebase()
    this.ref = this.firebase.child('accum')
    this.ref.on('value', () => 2)
  }
}
```

this will throw an error because these properties are not serializable.

The current solution? To either explicitly specify what is serializable via lifecycle methods, or use symbols to store that data. While this is nice and explicit I think we can remove this boilerplate entirely.

Cons of this change:

* It's a breaking change and that stinks.
* It makes the API very react-like.
* It is probably not the JS we will be writing in the future if JS does end up with private instance properties.

Proposal for migrating to this:

* [ ] Introduce warning in next release
* [ ] Change all examples to this new syntax
* [ ] Make this backwards compatible somehow
* [ ] Remove old behavior in 1.0?
* [ ] Write a script that will automatically migrate you from old to new.

frederickfogerty:
Yes, I love it. I agree with all the points you have raised. Other Flux stores don't have this issue as they force users to implement their own public methods, whereas alt relies on `getState`. In other Flux stores thus you can have private and public properties without much effort. This would be the best way to do it in alt.

It's a yes from me.

troutowicz:
:+1:

srph:
* insert traditional thumbsup + sparkle * :thumbsup: :sparkles:

[...]

PR #173, opened 2015-04-17
Allow customizing your state property key

goatslacker:
Lets have our :cake: and eat it too.

This PR adds a way out of #158 where we can get the best of both worlds. Through alt's config object you can specify a state key where your state will live:

```js
const alt = new Alt({ stateKey: 'bears' })

class Foo {
  static displayName: 'FooStore'

  constructor() {
    this.lookThisIsPrivate = 'yes it is'

    this.bears = {
      state1: 1,
      state2: 2
    }
  }
}
```

### DIFF
--- a/src/alt.js
+++ b/src/alt.js
@@ -445,7 +445,14 @@ function createStoreFromClass(alt, StoreModel, key, ...argsForConstructor) {
   const store = new Store(...argsForConstructor)
 
   storeInstance = assign(
-    new AltStore(alt.dispatcher, store, null, StoreModel),
+    new AltStore(
+      alt.dispatcher,
+      store,
+      typeof alt._stateKey === 'string'
+        ? store[alt._stateKey]
+        : null,
+      StoreModel
+    ),
     getInternalMethods(StoreModel, builtIns)
   )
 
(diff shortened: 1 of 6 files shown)


src/alt.js
```diff
@@ -454,6 +461,7 @@ function createStoreFromClass(alt, StoreModel, key, ...argsForConstructor) {
 
 class Alt {
   constructor(config = {}) {
+    this.config = config
```
jdlehman: I think it makes more sense to store `this.stateKey = config.stateKey` rather than the entire config.

jdlehman:
Great idea to make everyone happy. The config is paying off big time :smile:

src/alt.js
goatslacker: I figure I would need the entire config at some point in the future, no?. Should I put it in a symbol?

jdlehman: I like the pattern of being explicit and only the constructor getting the config. Then each thing that we want alt to know about can be pulled off the config. That is just my preference, I think either way is fine.  If we do go this way, I don't think we need a symbol, it would be fine as is.

PR #177, opened 2015-04-18
Implements connectToStores HoC

goatslacker:
#115


cpsubrian:
Thanks for taking over and getting this in! Got wrapped up with some PHP client work last week. Awesome that there is now more ES6 class support :)

goatslacker:
Sorry again that it took so long. The tests just needed a bit of massaging with all the new refactors. Thanks for all your hard work on this PR it was great!

Issue #171, opened 2015-04-17
Convert the utils to ES6


goatslacker:
We can put them in src/utils and then transpile them into utils.

PR #179, opened 2015-04-18
Convert utils to ES6

jdlehman:
- move utils to `src/utils`
- convert to ES6
- resolves Issue #171


components/AltNativeContainer.js
```diff
@@ -7,7 +7,6 @@ var React = require('react-native')
 var mixinContainer = require('./mixinContainer')
 var assign = require('object-assign')
 
-var View = React.View
```
goatslacker: ?

goatslacker: Oh I see what's going on. This is broken :( we need to pass in View into this.altRender and React.DOM.div for the other one.

utils/ActionListeners.js
```diff
@@ -1,62 +0,0 @@
-'use strict'
```
goatslacker: I was hoping we'd keep these here so we don't break linking:

`require('alt/utils/Util')`

vs

`require('alt/lib/utils/Util')`

goatslacker: The utils folder would need to be .gitignored though.

goatslacker:
This is awesome, thanks :+1: 

I converted some of the utils from class form to prototypes because they take up way less code that way.

jdlehman:
Cool by me!

PR #180, opened 2015-04-18
Add first class ImmutableJS support into alt

goatslacker:
ref: #97 

cc @ziflex @jdlehman 

Thanks to @troutowicz for his "make alt more pluggable" comment.

Check the tests for the API.

This needs docs.


jdlehman:
Glad all of our discussion led to this, much improved over the last iteration.  :+1: once this has docs

src/utils/ImmutableUtil.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  store.lifecycle.deserialize = function (obj) {
+    return Immutable.fromJS(obj)
+  }
+
+  return store
+}
+
+function makeImmutableClass(alt, Store, iden) {
```
jdlehman: do you need alt as an arg? you do not appear to be using it.

goatslacker: good catch

docs/immutable/immutablejs.md
```diff
@@ -1,163 +0,0 @@
-# Using Immutable Data Structures
```
goatslacker: should I move this to a section below on manually using ImmutableJS?

jdlehman: I think that would be a good idea. The util is what we recommend, but some people might have reasons to do it manually.  Plus I think it does a good job as an example showing how you can hook into serialization etc.  Might give people ideas to integrate other immutable libs themselves.

tldr; never hurts to have more docs :smile:

ziflex:
So, all alt extensions will be added via 'utils' ?

goatslacker:
`createStore` signature will remain as

> Store, identifier, ...args

where args are the arguments you want to inject into the constructor. I don't want to start adding more parameters to that function since it'll become a mess :) I have thought about store-level config object but I don't know just yet.

At the moment yeah all alt extensions added via utils.

And what do you mean by immutable store working with mutable data?

ziflex:
Actually, it makes sense to leave current signature. But, maybe rename 'utils' to 'extensions' or 'enhancers' to make their purpose more clear ?

> And what do you mean by immutable store working with mutable data?
As I can see that overridden method checks whether the state is __immutable__ and if not so fallback to original one. I'm not sure, that it's clear behavior. 
Why not like immutable stores are only for immutable data structures?

goatslacker:
Oh that's for the other stores. I'm overriding setState and getState for the entire alt instance so I need to fallback to default behavior if the store isn't immutable.

sogko:
This is great! Looking forward to using this.

I have yet to dive deep into this but I have a quick question: is ```createImmutableStore()``` the immutable equivalent of ```createStore()``` or ```addStore()```?

I noticed in the [current docs](https://github.com/goatslacker/alt/blob/4a70f8e422591080d39552f67b0973c800ae1a11/docs/utils/immutable.md), ```createImmutableStore()``` is called for ```alt``` instances, similar to  ```addStore()``` (while ```createStore()``` is called from exported ```alt``` from ```required('alt')```).

Thanks guys!

PR #181, opened 2015-04-18
Update README.md

indus:
Would you mind to add this badge?


goatslacker:
Don't mind at all. Thanks for the PR and for the domain

PR #186, opened 2015-04-19
Fix typo in comments

kyriacos:
FooStore != FootStore :)


goatslacker:
Thanks, I was building a Shoe App apparently.

PR #194, opened 2015-04-22
Add all actions to this.actions

### LINKED ISSUE #193: Add actions to this.actions

goatslacker:
Fixes #193 

This adds all the actions into `alt.actions` the actions are added individually rather than grouped this is so generateActions as well as single actions play well. This also results in the least amount of code. If this is awful though we can change it.


jareware:
Does this mean an action that previously was available as `myAlt.actions.someNs.myAction` can now be called directly as `myAlt.actions.myAction`?

goatslacker:
Namespaced actions will exist in their proper namespace. Actions that don't have a name will be listed under global.

goatslacker:
I think this is the last thing before 0.16.0 @jdlehman thoughts? Anything else we should include in 0.16.0

I've changed this a bit, I'll update with latest soon.

jdlehman:
This looks good to me. I cannot think of anything else that needs to be included in 0.16.0.  We could deprecate "magic static methods" in stores, but we could also do that later (if we even do it at all).

goatslacker:
Ok so here are the new changes:

* POJO actions can now have a displayName prop

```js
const pojo = alt.createActions({
  displayName: 'ActionName',

  one() { },
  two() { }
})
```

* Things created with alt.generateActions go into the `single` bucket

```js
alt.generateActions('one'); // alt.actions.single.one
```

* If there are name conflicts a number is suffixed

```js
alt.generateActions('one');
alt.generateActions('one');
// alt.actions.single.one, alt.actions.single.one1
```

* Everything else is namespaced

```js
@createActions(alt)
class Foo {
  static displayName: 'Foo',

  one() { }
}

alt.actions.Foo.one
```

* Unless we don't know the name

```js
alt.createActions(function () { this.one = function () { } })
// alt.actions.Unknown.one
```

src/alt/index.js
```diff
@@ -138,10 +145,28 @@ class Alt {
...
-      obj[action] = this.createAction(`${key}#${action}`, actions[action], obj)
+      if (typeof actions[action] !== 'function') {
+        return obj
+      }
+
+      // create the action
+      const name = `${key}#${action}`
+      obj[action] = this.createAction(name, actions[action], obj)
+
+      // remove the action referenced in single
```
jdlehman: This feels a bit round about to me.  The fact that creating the action always puts it on the `single` property seems inflexible. Maybe we can pass in the key on actions that we want to attach it to?

goatslacker: Ugh. I had the same thought but I thought maybe this hack was passable. I'll opt for that instead.

goatslacker: Also, care for the name single? I tried `global` but the actions aren't necessarily global, well maybe they are idk.

jdlehman: `single` is fine with me.  Nothing better comes to mind at the moment.

goatslacker: global feels right now that I think about it again. It's global to your alt instance.

justjake:
Sounds good to me. How often do people use one-off actions with `alt.generateActions`?

goatslacker:
I use `generateActions` quite a bit since it's easier to write than `this.generateActions` inside of a class. This also works well if you're not doing any async work.

PR #195, opened 2015-04-22
Adds a prepare for bootstrap method to stores

goatslacker:
Just a thought, we can call this function and have it return to us a serialized string ready for bootstrap. Could be useful if you have a chunk of data you want to load into a store.


jdlehman:
This is a nice helper function. :+1:

PR #204, opened 2015-04-24
Immutable decorator

goatslacker:
Playing with this API, I think it's way nicer than the current one.

Thoughts @ziflex @jdlehman


goatslacker:
Sorry the API is here: https://github.com/goatslacker/alt/blob/immutable-decorator/test/immutable-stores.js#L13

```js
@immutable
class TodoStore {
  constructor() {
    this.state = Immutable.Map({})
  }
}
```

ziflex:
Hell yeah! That's perfect solution!

jdlehman:
That is much better!

PR #208, opened 2015-04-25
Add a time travel decorator

goatslacker:
#55 

Decorator for allowing time traveling on stores

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+src/utils/TimeTravel.js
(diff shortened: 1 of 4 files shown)


goatslacker:
An API that may work out:

```js
@timetravel;
class TodoStore {
  static displayName = 'TodoStore'

  constructor() {
    this.todos = {};
    this.selectedTodo = null;
  }
}

export default alt.createStore(TodoStore);
```

```js
import TodoStore from '../stores/TodoStore';
TodoStore.revertTo(id); // revert to a specific ID
TodoStore.rollback(3); // rollback 3 transactions
TodoStore.rollback({
  todos: 3,
  selectedTodo: 5
}); // rollback the todos state value 3 transactions but selectedTodo 5 (this can get a little weird)
```

jdlehman:
To ensure that memory usage is not a problem, we should allow the user to set a limit of the number of items to store in memory, and perhaps provide a default number of items that are stored.

jdlehman:
:clap:

PR #209, opened 2015-04-25
Add `transform` function to AltContainer

caasi:
with it's docs.

### DIFF
--- a/test/store-listener-component-test.js
+++ b/test/store-listener-component-test.js
@@ -40,6 +40,26 @@ const Store2 = alt.createStore({
   }
 })
 
+const transformMock = function transformMock(it) {
+  /**
+   * Due to the behavior of React.render on multiple children, one can't rely on
+   * the count of the mock.
+   *
+   * In the following case, React will render the node as a node with a chlid:
+   * `<div />`, then render it as a node with a child: `<span />`, then render
+   * it as a node with three children: `<span />`, `<strong />` and `<em />`.
+   *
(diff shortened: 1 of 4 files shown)


goatslacker:
This just needs tests and it's good to go.

caasi:
I am working on it.

test/store-listener-component-test.js
```diff
@@ -203,6 +255,18 @@ export default {
       assert(span.props.TestStore.x === 'foobar')
     },
 
+    'children props work with the transform function'() {
+      const node = TestUtils.renderIntoDocument(
+        <AltContainer stores={{ TestStore }} transform={transformMock}>
```
goatslacker: what are you trying to test for here?

caasi: I want to test the transform function has been called and get the same TestStore as I pass to it.

But I don't want to test it for every possibilities about props because all the other tests about props are passed which means the default transform function is ok.

What should I do if I want to make it more clear?

goatslacker: i don't think you need the transformMock function at the top for this, just pass in test store and then check for it in the children props, no?

caasi: Yes, you are right. I removed the unnecessary test.

goatslacker:
Thanks, this is neat

srph:
Really neat!

caasi:
:rainbow:

PR #210, opened 2015-04-26
Allow you generate a single action

goatslacker:
This is like `createAction` except that the definition is just a straight dispatch. Is it worth the 7 lines it adds?

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -92,6 +92,12 @@ class Alt {
     })
   }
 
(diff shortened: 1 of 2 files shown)


jdlehman:
Can't you do the same thing with `this.generateActions('singeAction')` if you only specify a single action?  Not sure if this adds more confusion then it helps.

goatslacker:
Yes. Derp

Issue #198, opened 2015-04-23
Pull deserialize and serialize out of lifecycle


goatslacker:
I'm not sure where they should exist but not on the lifecycle

jdlehman:
I agree with this.  Deserialize and serialize are strange as they are called as result to events in the lifecycle, but are not really part of the lifecycle themselves.

goatslacker:
I'm thinking this can live in the store configs.

PR #213, opened 2015-04-26
Move serialize/deserialize to store config

jdlehman:
This moves serialize/deserialize out of the lifecycle and to the store config.

Issue #198

### DIFF
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -15,9 +15,9 @@ In the example below, we will show how this technique can be used to `onSerializ
 
 `onSerialize` provides a hook to transform the store data (via an optional return value) to be saved to an alt snapshot. If a return value is provided than it becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `onSerialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`. If there is no return value, the default, [`MyStore#getState()`](stores.md#storegetstate) is used for the snapshot data.
 
-## DeonSerialize
+## onDeserialize
 
-`onDeserialize` provides a hook to transform snapshot/bootstrap data into a form acceptable to the store for use within the application. The return value of this function becomes the state of the store. If there is no return value, the state of the store from the snapshot is used verbatim. For example, if the `onDeserialize` received the data `{queryParams: 'val=2&val2=23'}`, we might transform the data into `{val: 2, val2: 23}` and return it to set the store data such that `myStore.val === 2` and `myStore.val2 === 23`. DeonSerialize can be useful for converting data from an external source such as a JSON API into the format the store expects.
+`onDeserialize` provides a hook to transform snapshot/bootstrap data into a form acceptable to the store for use within the application. The return value of this function becomes the state of the store. If there is no return value, the state of the store from the snapshot is used verbatim. For example, if the `onDeserialize` received the data `{queryParams: 'val=2&val2=23'}`, we might transform the data into `{val: 2, val2: 23}` and return it to set the store data such that `myStore.val === 2` and `myStore.val2 === 23`. onDeserialize can be useful for converting data from an external source such as a JSON API into the format the store expects.
 
 ## Example
 
--- a/docs/serialization.md
+++ b/docs/serialization.md
@@ -7,17 +7,17 @@ permalink: /docs/serialization/
 
 # Serialization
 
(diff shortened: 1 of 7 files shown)


jdlehman:
Still need to get the tests passing.  Looks like there might also be a problem with the store config object from prior commits.  There is a a Store that does not have the config set correctly.  Looks like it is coming from `store-model-test.js` and the store is set as:

```js
{ boundListeners: [ 'hello' ],
  StoreModel:
   { state: { x: 0 },
     bindListeners: { hello: [Object] },
     hello: [Function: hello],
     config: { getState: [Function: getState], setState: [Function: assign] } },
  dispatchToken: 'ID_41',
  displayName: '' }
```

which shows it is missing its config obj (it is on the StoreModel, but does not get set to the store correctly like the other alt Stores)

test/index.js
```diff
@@ -208,6 +208,15 @@ class SecondStore {
 const secondStore = alt.createStore(SecondStore, 'AltSecondStore')
 
 class LifeCycleStore {
+  static config = {
+    serialize: (state) => {
+      state.serialized = true
```
goatslacker: return from here?

goatslacker:
Fixed the tests, we were just looking in the wrong place for the config. There's one more test that needs fixing and it's the app-level serialization/deserialization configuration.

What's currently happening:

* When a store is created it uses the app-level configuration as a base: https://github.com/goatslacker/alt/blob/master/src/alt/index.js#L65
* This is so we can override getState, setState for every store without having to pass it individually to each store: https://github.com/goatslacker/alt/blob/master/src/alt/utils/createStoreConfig.js#L3

The problem is that we now have serialize/deserialize hooks that determine how a store should be serialized/deserialized. But we also have app-level serialize/deserialize if you want to use YAML instead of JSON for example.

Solutions?

test/index.js
jdlehman: whoops haha

jdlehman:
Hmm I did notice that and thought it might end up being a problem.  I guess we could just rename one pair.  I think serialize/deserialize makes sense at the app-level, as that is what JSON.stringify/JSON.parse are doing.  At the model level, it is pretty much selecting or setting up the data it wants to serialize/deserialize.  We could call them `preSerialize` or `serializePrepare` or `dataForSerialization`?

Just ideas off the top of my head.

jdlehman:
Yep, that fixes them. Once we settle on a name, I can get this up and ready to merge.

goatslacker:
how about onSerialize and onDeserialize?

jdlehman:
Oh, I did `serializeData` and `deserializeData` just to make sure it worked. I will s/serializeData/onSerialize and we can merge.

docs/serialization.md
```diff
@@ -7,17 +7,17 @@ permalink: /docs/serialization/
...
 
-## Serialize
+## onSerialize
 
-`serialize` provides a hook to transform the store data (via an optional return value) to be saved to an alt snapshot. If a return value is provided than it becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `serialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`. If there is no return value, the default, [`MyStore#getState()`](stores.md#storegetstate) is used for the snapshot data.
+`onSerialize` provides a hook to transform the store data (via an optional return value) to be saved to an alt snapshot. If a return value is provided than it becomes the value of the store in the snapshot. For example, if the store name was `MyStore` and `onSerialize` returned `{firstName: 'Cereal', lastName: 'Eyes'}`, the snapshot would contain the data `{...'MyStore': {'firstName': 'Cereal', 'lastName': 'Eyes'}...}`. If there is no return value, the default, [`MyStore#getState()`](stores.md#storegetstate) is used for the snapshot data.
 
-## Deserialize
+## DeonSerialize
```
goatslacker: :fire:

jdlehman: NOOOO :sob:

src/utils/ImmutableUtil.js
```diff
@@ -48,6 +30,14 @@ function immutable(store) {
 
     getState(currentState) {
       return currentState
+    },
+
+    onSerialize: (state) => {
```
goatslacker: either lets make these all arrow functions or lets keep them as method shorthand.

I vote for arrow:

```js
getState: state => state,
onSerialize: state => state.toJS(),
...
```

goatslacker:
:+1:

docs/serialization.md
jdlehman: Realized after that file that I needed to substitute deserialize first, but I missed fixing the mistakes with the capital D. #vimproblems

PR #214, opened 2015-04-27
Fixes immutable util for subsequent mutations

### LINKED ISSUE #212: Nested immutable structures in stores?
Hi,

 The recent updates re: immutable structures  got me really excited and decided to give it a try (although I understand it has not been tagged a release yet)

I've build my own copy and either found a bug or I might have misunderstood the intended design for the immutable stores.

Given an immutable state (Map) with a nested immutable Map and a nested immutable List,
it seems like calling ```setState``` would reset the ```this.state```, instead of updating it?

Is the following approach to nested immutable structures wrong?

Below is a quick test that I wrote. 

Thanks for your time!

```es6
// test/immutable-stores.js
  ...
    'using nested immutable structures'() {
      const alt = new Alt()

      const actions = alt.generateActions('replaceMap', 'replaceList', 'updateBar')

      @immutable
      class ImmutableStore {
        constructor() {
          this.bindListeners({
            handleUpdateBar: actions.updateBar,
            handleReplaceMap: actions.replaceMap,
            handleReplaceList: actions.replaceList
          })

          this.state = Immutable.Map({
            bar: 'hello',
            list: Immutable.List([]),
            map: Immutable.Map({})
          })
        }

        handleUpdateBar(x) {
          this.setState(this.state.set('bar', x))
        }

        handleReplaceMap(x) {
          this.setState(this.state.set('map', Immutable.fromJS(x).toMap()))
        }

        handleReplaceList(x) {
          this.setState(this.state.set('list', Immutable.fromJS(x).toList()))
        }
      }

      const store = alt.createStore(ImmutableStore, 'ImmutableStore')

      assert(store.getState().get('bar') === 'hello', 'bar is initially `hello`')
      assert(store.getState().get('map').count() === 0, 'map is initially zero')
      assert(store.getState().get('list').count() === 0, 'list is initially zero')

      actions.replaceMap({a: 1, b: 2})
      assert(store.getState().get('bar') === 'hello', 'bar is still `hello` after replacing map')
      assert(store.getState().get('list').count() === 0, 'list still has zero items after replacing map')
      assert(store.getState().get('map').count() === 2, 'map has 2 items in it now after replacing map')

      actions.replaceList([1, 2, 3, 4])
      assert(store.getState().get('bar') === 'hello', 'bar is still `hello` after replacing list')
      assert(store.getState().get('list').count() === 4, 'list has 4 items now after replacing list') 
      assert(store.getState().get('map').count() === 2, 'map still has 2 items in it after replacing list') // <--- FAILED, count() => 0

      actions.updateBar('world')
      assert(store.getState().get('bar') === 'world', 'bar is now `world` after updating bar')
      assert(store.getState().get('list').count() === 4, 'list still has 4 items  after updating bar') // <--- FAILED, count() => 0
      assert(store.getState().get('map').count() === 2, 'map still has 2 items in it after updating bar') // <--- FAILED, count() => 0
    }
...
```

goatslacker:
Fixes #212 

The problem was that `this.state` was referencing the original Immutable object, so any subsequent mutations would result in the original object + whatever mutations you just made. This means that if you did five consecutive mutations only the first and last would apply.

The solution was to pass in the context of `this` to setState and getState so I'm able to poke into the store and update `this.state` so you're able to re-use it.

I've added both test cases from #212 as well.

### DIFF
--- a/dist/alt-with-addons.js
+++ b/dist/alt-with-addons.js
@@ -1099,6 +1099,7 @@ var AltStore = (function () {
     this[LIFECYCLE] = {};
     this[STATE_CONTAINER] = state || model;
 
+    this._storeName = model._storeName;
     this.boundListeners = model[ALL_LISTENERS];
     this.StoreModel = StoreModel;
     if (typeof this.StoreModel === 'object') {
@@ -1173,7 +1174,7 @@ var AltStore = (function () {
   }, {
     key: 'getState',
     value: function getState() {
-      return this.StoreModel.config.getState(this[STATE_CONTAINER]);
+      return this.StoreModel.config.getState.call(this, this[STATE_CONTAINER]);
     }
   }]);
 
@@ -1544,6 +1545,16 @@ var Alt = (function () {
       });
     }
   }, {
+    key: 'prepare',
+    value: function prepare(store, payload) {
+      var data = {};
+      if (!store._storeName) {
+        throw new ReferenceError('Store provided does not have a name');
+      }
+      data[store._storeName] = payload;
+      return this.serialize(data);
+    }
+  }, {
     key: 'addActions',
 
     // Instance type methods for injecting alt into your application as context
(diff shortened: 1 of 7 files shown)


jdlehman:
:+1:

PR #220, opened 2015-04-30
Single store actions

goatslacker:
This refactors the files around again this time all the files pertinent to their domain are contained in a single file so they're not dispersed everywhere.

This reduces the min build by 4kb and the gzip build by 200bytes. This is not really significant but I sort of like the structure.

@jdlehman any preference?


jdlehman:
Structuring by domain makes sense to me, but I think with this refactor we are already reaching the limit where some of these files are not as approachable because they are so large.

I think I would prefer to have folders for domains and broken up files inside of them, just my preference though so take that with a grain of salt.  Obviously there are concerns about build size, but the amount of bytes that we are saving sounds like micro-performance to me.  Can we investigating deduping files via google closure compiler or some other tool?  That would give us the ability to break up the files into logical units as well as save on the build size.

I know webpack has a [deduping](https://github.com/webpack/docs/wiki/optimization#deduplication) plugin.  I can try to throw together a webpack build to see if I can get the file size down as well.

goatslacker:
??

```
alt/actions
  - Action.js
  - makeAction.js
alt/store
  - Store.js
  - StoreMixins.js
  - StoreUtils.js
```

jdlehman:
Yeah, I was thinking something akin to that.  But I am fine with whatever makes sense to most people.

goatslacker:
Switching out browserify with webpack would be cool for the browser build. Perhaps we can provide a pre-minned and optimized build as well. Uglify does a really good job at this.

goatslacker:
Ok yeah I think this is nicer.

goatslacker:
Speak now or forever hold your peace. Well, not forever :P

jdlehman:
:+1:

PR #221, opened 2015-04-30
Fix altmanager createstore call

chetstone:
Fix altmanagermixin to conform to new breaking changes in the call signature of createStore in 0.16.1. 
Also includes "bonus" fix to initially generate the correct weather for SF, CA.


goatslacker:
Sweet, thanks!

PR #222, opened 2015-04-30
Use webpack for browser builds

jdlehman:


jdlehman:
@goatslacker here is a quick webpack config I through together, it probably still needs to be optimized some. Strangely it did not seem to help size at all.  I know like Babel, webpack also adds some helpers it uses so maybe that is adding extra weight?  I am no webpack expert so I could also be missing some configs.

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+  module: {
+    loaders: [
+      { test: /\.js$/, loader: 'babel-loader', exclude: /node_modules/ }
+    ]
+  },
+  output: {
+    path: __dirname + '/dist',
+    filename: getOutputFilename(production),
+    libraryTarget: 'umd',
+    library: 'alt'
```
goatslacker: does this mean it'll be available on `window.alt`?

webpack.config.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+  externals: [
+    {
+      "react": {
```
goatslacker: it's not excluding react

goatslacker:
this is a good starting point

webpack.config.js
troutowicz: I've always had a difficult time following webpacks documentation... but this works for me:

```js
externals: [ /^react(\/.*)?$/ ]
```

webpack.config.js
jdlehman: This exports the whole bundle as a library so a user could just do `import 'alt';`

http://webpack.github.io/docs/configuration.html#output-library

webpack.config.js
jdlehman: Oh probably because it isn't excluding the addons. Thanks @troutowicz, I'll fix that.

david0178418:
This will be great as it seems that this won't require React to be global as it does now.

PR #224, opened 2015-05-01
export flux constructor so user does not need require it outside again

davidxi:
I'm using `alt.js` to help me rewrite my existing code. However, I have used `flux` for other purposes in the existing code. So i'm wondering if I used the bundle version of `alt.js`, is it ok it also export the `flux` constructor?

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -220,4 +220,5 @@ class Alt {
   }
 }
 
+export { Dispatcher }
 export default Alt


goatslacker:
Yeah this seems like an ok change. Maybe we can add it to Alt as a static property?

What's the use case for this?

davidxi:
one use case from my experience is that I used `flux` also as a callback controller in my code.

goatslacker:
As a callback controller? How so?

davidxi:
sorry for confusion. i mean, like to use `waitFor()` or to trigger callbacks without touching dom

goatslacker:
Do you have some code samples of what it is you're doing?

You can access an instance of the dispatcher through alt:

```js
const alt = new Alt();

const dispatcher = alt.dispatcher;
```

This allows you to register your own callbacks:

```js
alt.dispatcher.register(payload => console.log(payload));
```

and you can also use `waitFor` from the dispatcher:

```js
alt.dispatcher.register((payload) => {
  alt.dispatcher.waitFor([MyStore.dispatchToken]);
});
```

or you can use it inside your stores:

```js
class OtherStore {
  constructor() {
    this.bindListeners({ handleFoo: MyActions.updateFoo });
  }

  handleFoo() {
    this.waitFor(MyStore);
  }
}
```

Furthermore, if you don't like Flux's dispatcher you can provide your own:

```js
const alt = new Alt({
  dispatcher: MyCustomDispatcher
});
```

Is there something that's not handled by these cases or that should be enhanced?

davidxi:
ah yes you are right. Actually I can just initialize an Alt instance anyway, and just use its `dispatcher` property. Because in my case, I also used flux dispatcher instance in non-react part, where neither action and store is needed.  On the other side, I saw there are no function calls  in Alt constructor. So, initializing for just a dispatcher should be not a big deal.

PR #226, opened 2015-05-01
Pass action details through the dispatch

### LINKED ISSUE #192: Action name available during dispatch
Having the action name available during dispatch would be useful for several reasons, for instance:
 * debugging to see which action is dispatching
 * sending actions over to an analytics tool

Currently the easiest way to get it is:
```js
app.dispatcher.register(function(payload) {
  var actionName = payload.action.toString().replace(/.*#(\w+).*/, '$1');
});
```
but digging it out from the Symbol through `toString()` isn't too pretty.

Also, when munging for production with e.g. `uglify` you lose at least the store name. Extra credit for figuring out how to preserve that as well.

goatslacker:
Fixes #192 we pass in a whole slew of details through the dispatch.

For this action class:

```js
class MyActions {
  static displayName = 'MyActions'

  updateName(name) { this.dispatch(name) }
}
```

we would get this info:

```js
details: {
  namespace: 'MyActions',
  name: 'updateName',
  id: 'MyActions.updateName',
  symbol: Symbol('alt/MyActions.updateName')
}
```

you can extra it via `payload.details`.

### DIFF
--- a/src/alt/AltAction.js
+++ b/src/alt/AltAction.js
@@ -3,14 +3,15 @@ import * as Sym from './symbols/symbols'
 const { ACTION_HANDLER, ACTION_UID } = Sym
 
 export default class AltAction {
-  constructor(alt, name, action, actions) {
+  constructor(alt, name, action, actions, actionDetails) {
     this[ACTION_UID] = name
     this[ACTION_HANDLER] = action.bind(this)
+    this.actionDetails = actionDetails
     this.actions = actions
     this.alt = alt
   }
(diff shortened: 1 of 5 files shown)


jdlehman:
Nice and simple :+1:

goatslacker:
cc @jareware @spikebrehm

jareware:
This looks delicious, :+1: !

spikebrehm:
Thank you! :+1:

PR #227, opened 2015-05-03
Configure eslint

jdlehman:
- Move to separate eslintrc file
- Based on airbnb's js [style guide](https://github.com/airbnb/javascript)
- Fixes new linting errors


jdlehman:
Issue #227

.eslintrc
```diff
@@ -0,0 +1,172 @@
...
+        "after": true
+      }],
+      "space-after-keywords": 2,
+      "space-before-blocks": 2,
+      "space-before-function-paren": [2, "never"],
+      "space-infix-ops": 2,
+      "space-return-throw-case": 2,
+      "spaced-line-comment": 2,
+
+    /* custom styles */
```
jdlehman: This commit is a good starting point. Not sure how we want to format these extra rules we already had prior to copying rules from the airbnb style guide.  Maybe this is fine with just a "custom rules" comment

.eslintrc
```diff
@@ -0,0 +1,172 @@
...
+        "after": true
+      }],
+      "space-after-keywords": 2,
+      "space-before-blocks": 2,
+      "space-before-function-paren": [2, "never"],
+      "space-infix-ops": 2,
+      "space-return-throw-case": 2,
+      "spaced-line-comment": 2,
+
+    /* custom rules */
```
jdlehman: This commit is a good starting point. Not sure how we want to format these extra rules we already had prior to copying rules from the airbnb style guide. Maybe this is fine with just a "custom rules" comment

src/utils/makeFinalStore.js
```diff
@@ -25,7 +25,8 @@
 function FinalStore() {
   this.dispatcher.register((payload) => {
     const stores = Object.keys(this.alt.stores).reduce((arr, store) => {
-      return arr.push(this.alt.stores[store].dispatchToken), arr
```
goatslacker: +1

src/utils/withAltContext.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 export default function withAltContext(flux) {
-  return function (Component) {
+  return function(Component) {
```
goatslacker: Can we revert these? My take on it is that function is a keyword similar to if, while, etc and should be treated as such in regards to spacing.

goatslacker:
:+1:

src/utils/withAltContext.js
jdlehman: We can do whichever.  Right now we have some instances with the space and some without. So I can add the rule so that we require the space so that we can be consistent everywhere.

jdlehman: Do we want to have a space for named functions too or just anonymous functions?

Like in ES6:
```js
myFunction(args) {
}
```

or 
```js
myFunction (args) {
}
```

PR #228, opened 2015-05-04
Make our own assign

goatslacker:
object-assign uses the Object.assign polyfill if available. react-native has its own stupid polyfill which. We are abusing Object.assign sometimes and hitting some of the "edge cases" that react-native's polyfill doesn't support for perf reasons. So lets just do forego the Object.assign polyfill and have our own. Bonus points: it makes our build smaller.

### DIFF
--- a/components/AltContainer.js
+++ b/components/AltContainer.js
@@ -56,7 +56,7 @@
  */
 var React = require('react/addons')
 var mixinContainer = require('./mixinContainer')
-var assign = require('object-assign')
+var assign = require('../utils/functions').assign
 
 var AltContainer = React.createClass(assign({
   displayName: 'AltContainer',
--- a/components/AltNativeContainer.js
+++ b/components/AltNativeContainer.js
@@ -5,7 +5,7 @@
  */
 var React = require('react-native')
 var mixinContainer = require('./mixinContainer')
-var assign = require('object-assign')
+var assign = require('../utils/functions').assign
 
 var AltNativeContainer = React.createClass(assign({
   displayName: 'AltNativeContainer',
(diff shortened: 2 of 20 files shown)


goatslacker:
Thoughts on this?

jdlehman:
There are some tradeoffs.  We get a smaller build and can remove some of these edge cases, but we also have to maintain our own polyfill.  This could mean we could have issues with bugs and other edge cases popping up that might get to the point where we have to add enough code such that we don't gain anything on build size.  `object-assign` is an extremely popular lib, so it might even make sense to try to get these edge cases resolved there instead.

goatslacker:
actually the build isn't smaller it's slightly larger :(

I'm not really happy with this solution.

Amending `object-assign` is out of the question because this change isn't to fix an edge case but rather to hack around the fact that react-native's polyfill isn't spec compliant.

object-assign is meant to work by checking if `Object.assign` exists, if it does then it uses that, if it doesn't then it polyfills it. What I need is a polyfill that disregards Object.assign because react-native implements its own Object.assign

jdlehman:
ohhh. hmmm. we can make it an assign util and provide a config option to swap it out?

```js
var alt = new Alt({
   assign: ourObjectAssign
});
```

goatslacker:
hmmm I'm not sure how much value that adds as a whole assign is pretty straightforward and I don't know how much you can configure it. Plus it would bork all of internals.

jdlehman:
Oh good point.  I guess it is fine to move forward with this.  If it ends up not working for some reason it would not be difficult to go back.

goatslacker:
turns out the build is actually slightly smaller.

jdlehman:
Can we move this new assign into its own repo?  It feels like a better separation of concerns to me.  Then all we need to change in alt is the package and imports.  Though I have been thinking about this, and I think I am fine with it overall if it gives us some React Native wins.

goatslacker:
Yeah the only reason I'm considering this is because of react native. I could publish this as its own npm module as well. Idk. I'm going to merge now and fix later.

PR #231, opened 2015-05-04
Replace eventemitter3 with contra.emitter

goatslacker:
Just an experiment in reducing build size and complexity.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,6 @@
     // possible errors
     "comma-dangle": [2, "never"],
     "no-cond-assign": [2, "always"],
-    "no-console": 1,
     "no-debugger": 1,
     "no-alert": 1,
     "no-constant-condition": 1,
@@ -128,6 +127,7 @@
       "spaced-line-comment": 2,
 
     /* custom rules */
+    "no-console": 0,
     "space-before-function-paren": [2, {
       "named": "never",
       "anonymous": "always"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ after_script: cat ./coverage/lcov.info | coveralls
 node_js:
   - "0.10"
   - "0.12"
-  - iojs
+  - iojs-v1.8.1
--- a/dist/alt-with-addons.js
+++ b/dist/alt-with-addons.js
@@ -136,13 +136,12 @@ function mixinContainer(React) {
     },
 
     registerStores: function registerStores(props) {
+      var stores = props.stores;
       Subscribe.create(this);
 
       if (props.store) {
         this.addSubscription(props.store);
       } else if (props.stores) {
-        var stores = props.stores;
-
         if (Array.isArray(stores)) {
           stores.forEach(function (store) {
             this.addSubscription(store);
@@ -160,11 +159,10 @@ function mixinContainer(React) {
     },
 
     getStateFromStores: function getStateFromStores(props) {
+      var stores = props.stores;
       if (props.store) {
         return getStateFromStore(props.store, props);
       } else if (props.stores) {
-        var stores = props.stores;
-
         // If you pass in an array of stores then we are just listening to them
         // it should be an object then the state is added to the key specified
         if (!Array.isArray(stores)) {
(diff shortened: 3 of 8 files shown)


goatslacker:
This will have to land in 0.17.0 since it's a breaking change.

koistya:
But... web apps may use `eventemitter3` anyway, for some other needs. If so, it may result into two even emitter libraries bundled into app.js

goatslacker:
@koistya that is a very valid concern. I'd be interested in seeing if this is the case and if so we wouldn't make the change.

goatslacker:
Does it make sense to be tied to EE3 because of other people possibly using it? I'm using v0.1.6 of EventEmitter3 and the latest is v1.1.0

I have no intention of upgrading since it adds a function that we don't need. Users might be on v1 already or they may lag behind on v0.1.6 thus already getting different versions.

jdlehman:
I think this is fine.  We can't base our decisions off guesswork about what libraries users are using in conjunction with alt.  EE3 might be very popular, but if contra makes more sense for us, then I say go for it.

cmwelsh:
Partnerships and commitments are important. It's like saying Bluebird is the native promise lib for your project, which helps a certain percentage of people gravitate towards your project because they can increase the number of peerDependencies. Or you do what Reflux does and allow random libs to be configurable:

https://github.com/spoike/refluxjs/commits/master#switching-promise-library

goatslacker:
Yup that's pretty important, however these event emitters are pretty small where they may be negligible.

Reflux offers the illusion of being able to manage your dependencies, bluebird and EventEmitter still get required so you'll be incurring the bytes penalty from those libraries plus your own. It is nice that they offer that flexibility though.

I am interested in hearing more feedback like this though! Do you think it's worth it still if I don't ever plan on upgrading past 0.1.6?

mikekidder:
This is preliminary, but maybe you should just stay @0.6.1.  Using [primus\eventemitter3](https://github.com/primus/eventemitter3/tree/master/benchmarks) own benchmark testing (needs updating), I got this on my machine for the ```emit.js``` test.

> log: Finished benchmarking: "EventEmitter1"
 metric: Count (244504), Cycles (5), Elapsed (5.507), Hz (4733544.124308651)
 log: Finished benchmarking: "EventEmitter2"
 metric: Count (236391), Cycles (4), Elapsed (5.295), Hz (4539167.535433991)
 log: Finished benchmarking: "EventEmitter3@0.6.1"
 metric: Count (348439), Cycles (5), Elapsed (5.399), Hz (6249142.761644134)
 log: Finished benchmarking: "EventEmitter3(master)"
 metric: Count (74041), Cycles (7), Elapsed (5.802), Hz (1418540.8183131793)
 log: Finished benchmarking: "Drip"
 metric: Count (271637), Cycles (4), Elapsed (5.329), Hz (5123509.586292663)
 log: Finished benchmarking: "fastemitter"
 metric: Count (186534), Cycles (6), Elapsed (5.372), Hz (3605665.189683198)
 log: Finished benchmarking: "event-emitter"
 metric: Count (140395), Cycles (6), Elapsed (5.428), Hz (2715251.819991193)
 log: Finished benchmarking: "contra.emitter"
 metric: Count (7500), Cycles (6), Elapsed (5.465), Hz (136538.49571436027)
 info: Benchmark: "EventEmitter3@0.6.1" is the fastest.

As you can see, **eventemitter3@0.6.1** takes the cake and **eventemitter2@master** is slow.  But,  **contra.emitter** is somewhere in the dust.  

Looks like a PR request to there benchmarks is in order :)

goatslacker:
What if: **crazy idea** we implement our own event emitter?

It's not very complicated and I can make it faster than all of those because I only support a very specific use case.

Pros:

Really small build.

Cons:

Breaking change, I'll remove the getEventEmitter() method because

* It's an anti-pattern to be emitting events.
* It won't be a full event emitter any more.

The lifecycle listeners will now behave like this emitter.

af7:
I agree with the *crazy idea* (if it is doable). The smaller the build the better!

goatslacker:
If I remove event emitter and symbol polyfill the min build goes to 20kb. 5kb gzip.

koistya:
> What if: crazy idea we implement our own event emitter?

:+1:

goatslacker:
#248

PR #238, opened 2015-05-05
Improve consistency of action import on example

firstdoit:
From line 171 onwards, all Stores and Actions classes start with uppercase and all **created** actions and stores start with lowercase:

```
var locationActions = alt.createActions(LocationActions)
```

Therefore, in the first and foremost example, when we `export default alt.createActions(TodoActions);`, we should probably import it into a lowercase var.

Opinions on this?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,6 @@
     // possible errors
     "comma-dangle": [2, "never"],
     "no-cond-assign": [2, "always"],
-    "no-console": 1,
     "no-debugger": 1,
     "no-alert": 1,
     "no-constant-condition": 1,
@@ -128,6 +127,7 @@
       "spaced-line-comment": 2,
 
     /* custom rules */
+    "no-console": 0,
     "space-before-function-paren": [2, {
       "named": "never",
       "anonymous": "always"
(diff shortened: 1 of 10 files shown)


goatslacker:
I usually import them as uppercase because they are singleton objects, the examples at the bottom show them in lowercase because the uppercase name is already taken. So those classes should either change name or we should be consistent about this.

I don't really have an opinion though. The flux community seems to use PascalCase for stores and actions so I'm just following convention there.

firstdoit:
You have very good points. But I can't think of a better name for the classes that won't seem artificial.

goatslacker:
Yeah I'm not sure either other than anonymously creating the classes like:

```js
const LocationStore = alt.createStore(class {
  static displayName = 'LocationStore'

  constructor() {
  }

  ...
})
```

jdlehman:
You could go `LocationStoreClass` though that is pretty terrible... :disappointed:

firstdoit:
> There are only two hard things in Computer Science: cache invalidation and naming things.
> Phil Karlton

goatslacker:
I'll be going through a docs pass in 0.17.0 and will be revising this.

PR #247, opened 2015-05-11
Update connectToStores ref in docs

frederickfogerty:
I expect a lot of issues merging this

### DIFF
--- a/guides/getting-started/view.md
+++ b/guides/getting-started/view.md
@@ -33,7 +33,7 @@ componentWillUnmount() {
 },
 ```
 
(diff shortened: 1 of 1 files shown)


goatslacker:
Why the blob rather than the github view? The github view is there because it offers syntax highlighting for viewing and reading the source, you can always click raw to get the blob.

frederickfogerty:
Oh, I just copied and pasted from the url, didn't know there was a difference

goatslacker:
Oops, never mind my mistake!

goatslacker:
Thanks :D

PR #248, opened 2015-05-12
Remove EventEmitter3 and Symbol from Alt for a smaller build

goatslacker:
With this change we're now at 19kb min and 5kb gzip.


jdlehman:
Wow :clap:, this is heroic.

1. I can't believe transmitter was not taken on npm.
2. can you push transmitter to github so it is easier to take a look at (I am sure you probably want to get it into a more presentable state with docs/readme before you do so)

goatslacker:
Sounds good

goatslacker:
I have to detail exactly what this breaks. It's not much actually but I still want to have a comprehensive list.

troutowicz:
For stores just passing a model to the  `createStore` method, it looks like testing would be a breeze!

```js
cont UnwrappedStore = AppStore.state;
```

goatslacker:
you already have unwrapped store:

`const UnwrappedStore = AppStore.StoreModel` :)

but yes you now have access to state...

goatslacker:
https://github.com/goatslacker/transmitter

jdlehman:
I like the simplicity of transmitter.  This has my approval

goatslacker:
This is merged into #253

PR #251, opened 2015-05-14
Allow connectToStore components to use props in getStores

firstdoit:
This allows for: 

```
  statics: {
    getStores(props) {
      return [
        props.flux.stores.SearchStore
      ];
    },

    getPropsFromStores(props) {
      return props.flux.stores.SearchStore.getState();
    }
  }
```

### DIFF
--- a/src/utils/connectToStores.js
+++ b/src/utils/connectToStores.js
@@ -10,7 +10,7 @@
  *
  *    const MyComponent = React.createClass({
  *      statics: {
- *        getStores() {
+ *        getStores(props) {
  *          return [myStore]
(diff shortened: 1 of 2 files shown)


src/utils/connectToStores.js
```diff
@@ -65,12 +62,14 @@ function connectToStores(Component) {
     },
 
     componentDidMount() {
+      let stores = Component.getStores(this.props)
```
goatslacker: can you change this to const?

goatslacker:
Nice :+1:

PR #252, opened 2015-05-16
added helpful warnings for Flux instance methods 

asfktz:
if will warn in those cases:

this.addActions('fooActions', null);
this.addActions('fooActions', undefined);
=>  fooActions must be an Array or an ActionsClass

this.addStore('fooStore', null);
this.addStore('fooStore', undefined);
=>  fooStore must be a StoreModel

this.addActions(FooActions);
=> actions must have name

this.addStore(FooStore);
=> store must have name

getStore('fooStore')
=> "fooStore" not found. available stores are: productStore, cartStore

getActions('fooActions')
=> "fooActions" not found. available actions are: productActions, cartActions

also added console.trace() to utils.warn() to print a call stack

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -185,20 +185,48 @@ class Alt {
   // Instance type methods for injecting alt into your application as context
 
   addActions(name, ActionsClass, ...args) {
+    if (typeof name !== 'string') utils.warn(`actions must have name`)
+
+    if (Object.is(ActionsClass, undefined) ||
+        Object.is(ActionsClass, null)) {
+      utils.warn(`${name} must be an Array or an ActionsClass`)
+    }
+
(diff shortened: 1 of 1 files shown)


src/alt/index.js
```diff
@@ -185,20 +185,48 @@ class Alt {
   // Instance type methods for injecting alt into your application as context
 
   addActions(name, ActionsClass, ...args) {
+    if (typeof name !== 'string') utils.warn(`actions must have name`)
+
+    if (Object.is(ActionsClass, undefined) ||
```
goatslacker: We can't use Object.is since it won't be available to other browsers

goatslacker: can we configure this so that you can use envify to remove this from production builds?

asfktz:
ok, will make the change

Issue #257, opened 2015-05-20
Add shouldFetch to Sources


goatslacker:
It would return a Boolean.

`false` means it would never call `remote` even if `local` returns falsy value.

`true` means it always calls `remote`.

PR #260, opened 2015-05-21
Allow shouldFetch function in data source for async calls

jdlehman:
If shouldFetch returns true, it will always call remote to get the value, and if it returns false, it will always use the local value. If it is not defined, the presence of the local value determines if remote is called or not.

#257

### DIFF
--- a/src/alt/store/StoreMixin.js
+++ b/src/alt/store/StoreMixin.js
@@ -38,9 +38,10 @@ const StoreMixin = {
       publicMethods[methodName] = (...args) => {
         const state = this.getInstance().getState()
         const value = asyncSpec.local && asyncSpec.local(state, ...args)
+        const shouldFetch = asyncSpec.shouldFetch ? asyncSpec.shouldFetch(state, ...args) : !value
 
         // if we don't have it in cache then fetch it
-        if (!value) {
+        if (shouldFetch) {
           isLoading = true
           hasError = false
           /* istanbul ignore else */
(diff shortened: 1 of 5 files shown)


src/alt/store/StoreMixin.js
```diff
@@ -38,9 +38,10 @@ const StoreMixin = {
       publicMethods[methodName] = (...args) => {
         const state = this.getInstance().getState()
         const value = asyncSpec.local && asyncSpec.local(state, ...args)
+        const shouldFetch = typeof asyncSpec.shouldFetch === 'function' ? asyncSpec.shouldFetch(state, ...args) : !value
```
goatslacker: I think we should assume it's always a function if its defined.

goatslacker: this is clever though :)

goatslacker:
:+1:

src/alt/store/StoreMixin.js
jdlehman: I was initially doing a `typeof asyncSpec.shouldFetch === 'undefined'` check, but then since I had to explicitly check anyway, I figured I might as well check if it is a function and get a little added benefit.

We cannot do  `async.shouldFetch ? asyncSpec.shouldFetch : !value` because if `shouldFetch` is false it should still be used.

Would you like me to change it to check if it is undefined still?

goatslacker: You mean if shouldFetch: false. 

Why not just always make it a function?

shouldFetch: () => false

jdlehman: Oh right, I see what you are saying. :+1:

goatslacker:
npm run build and merge!

PR #264, opened 2015-05-22
Fix isLoading for multiple async calls

goatslacker:
Fixes isLoading call if multiple requests are in-flight.

Given the nature that the status is reset every time a new call is made a situation may arise where the isLoading boolean is prematurely set to false. Changing this to a counter fixes the issue.

cc @wlue


wlue:
This looks good to me. :+1:

PR #268, opened 2015-05-25
Update link to API Reference

iyn:
**What**
I updated the link to API reference from [https://goatslacker.github.io/alt/](https://goatslacker.github.io/alt/) to [http://alt.js.org/docs/](http://alt.js.org/docs/)

**Why**
The current address links to the site with broken styles (tested on Linux with FF 38 and Chromium 43 with and without extensions) - it seems that's the same page as [the official one](http://alt.js.org/) but without any styles (maybe that's the problem on my end?). But I guess it's also good to have consistent URLs in the project.

**Post Scriptum**
I love `alt` :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ There are also many [utils](/src/utils) available which interface well with alt:
 
 ## Topical Guide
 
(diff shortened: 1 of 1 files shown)


goatslacker:
Thanks!!!

PR #272, opened 2015-05-27
Update DateSource intercept to recive method args

chrishoage:
Pass in arguments that DataSource method was called with to intercept as an array


goatslacker:
Looks good, merging once build passes.

goatslacker:
Awesome, thanks, will put out a release soon, and will update the docs.

PR #278, opened 2015-05-29
Add batchingFunction support for Flux Dispatcher

chrishoage:
Adds support for a batchingFunction though optional config. To be used with React.addons.batchedUpdates

Still needs tests.

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -13,6 +13,7 @@ class Alt {
     this.serialize = config.serialize || JSON.stringify
     this.deserialize = config.deserialize || JSON.parse
     this.dispatcher = config.dispatcher || new Dispatcher()
+    this.batchingFunction = config.batchingFunction || (callback => callback())
     this.actions = { global: {} }
     this.stores = {}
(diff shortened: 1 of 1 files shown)


jdlehman:
Interesting idea, would be great to see an example of how you might use this.

goatslacker:
We were discussing this in the gitter channel, this is the gist of it:


1. A button in ComponentA calls an action creator
2. The action creator calls an async API
3. As a result of the async call, the action creator dispatches an action
4. That action sets new state in a store
5. The new store data causes a new child to be mounted inside ComponentA
   (let's call it ComponentB)
6. ComponentB fires an action immediately, via componentDid/WillMount

this usually causes a "cannot dispatch while dispatching" error because its being taken out of context from react's batching strategy

jdlehman:
:+1: thanks for the explanation.

chrishoage:
Thanks for finishing this up @goatslacker! Life and work got in the way over the weekend and couldn't find time to work on it.

goatslacker:
:+1: I'll put out a release today.

PR #279, opened 2015-05-30
Add FP goodies baked into alt

goatslacker:
Brings in a lot of the ideas of [microflux](https://github.com/goatslacker/microflux) into Alt.

This PR is full of new features:

* preventDefault() is this new method available on the store that will prevent a change event from being emitted. This is a better alternative than returning false from each action handler as this is more explicit. This will also be needed for other methods below.

* observe() is a new method that you define in your store. It was made mostly for when creating stores with POJOs and using alt instances. This method gets `alt` as its first and only parameter and you return an Object from it. The Object contains the same stuff that would go in `bindListeners`. These listeners are then bound.

* otherwise() is a new method you define in your store. This is a concept borrowed from [Haskell's guards](http://learnyouahaskell.com/syntax-in-functions#guards-guards). It is called for every single dispatch that isn't already being handled in your store. So say your store already has handlers for actions A and B and action C fires, this would hit the `otherwise` method. If action A were to be fired then only the handler would react.

* reduce() is a new method you define in your store. This handles *all* dispatches and it is quite a special method since it doesn't work like the other ones. In reduce you return an object which then becomes the next state. This works well if you're looking to create simple stores of derived data or as some call them: ["ephemeral stores"](https://github.com/appsforartists/funx). 

* output(), also a method you define in your store, shapes the data as its emitted in a change event. This is a method where you can transform the state (but not save) before it reaches the store listeners.

### DIFF
--- a/src/alt/store/AltStore.js
+++ b/src/alt/store/AltStore.js
@@ -13,41 +13,70 @@ class AltStore {
     this[Sym.LIFECYCLE] = model[Sym.LIFECYCLE]
     this[Sym.STATE_CONTAINER] = state || model
 
+    this.preventDefault = false
     this._storeName = model._storeName
     this.boundListeners = model[Sym.ALL_LISTENERS]
     this.StoreModel = StoreModel
+
+    const output = model.output || ((a, b) => b)
+
+    this.emitChange = () => {
+      this[EE].emit(
+        'change',
+        output(alt, this[Sym.STATE_CONTAINER])
+      )
+    }
+
+    const handleDispatch = (f, payload) => {
+      try {
+        return f()
+      } catch (e) {
+        if (model[Sym.HANDLING_ERRORS]) {
+          this[Sym.LIFECYCLE].emit(
+            'error',
+            e,
+            payload,
+            this[Sym.STATE_CONTAINER]
+          )
+          return false
+        } else {
+          throw e
+        }
+      }
+    }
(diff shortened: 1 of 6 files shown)


goatslacker:
And here's an example: https://gist.github.com/goatslacker/bdf5061f9b70a4fa4e99

jdlehman:
These are great!  I am especially excited about reduce.  I like how these are added in as features rather than requirements.  Let's people use alt how they need to.  Implementation looks solid from a first glance.

goatslacker:
A couple of API questions:

Should I pass in what action was called into reduce? What should the function arguments be?

What about otherwise, what should the signature be?

And output as well?

jdlehman:
I don't know if the action name is needed for reduce.  Any state they want to manage in the store should be handled accordingly based on the data passed in.  Some people might want it, but I think that might encourage large switch statements to handle logic rather than approaching it more functionally.  

I think the function arguments you have for `reduce` (alt, state, payload) are fine.

`otherwise` can just be the same as any other action handler (which I believe is how you have it now).

for `output` I think you just need the current state, which is also what you have now (along with the alt instance).

tldr; i think everything one would need as function arguments for these functions are there in the current implementation.

jdlehman:
We should add some docs for these :smile:

goatslacker:
on it!

PR #280, opened 2015-05-31
create .babelrc file and cleanup of scripts

mikekidder:
thought might be nicer to distinguish between build and transpile.
  
- removed "size" script as doesnt appear working/needed
- "prepublish" still runs the "transpile" twice as prior commit(s).
- added object-assign dep as used in "test-browser" script

essentially all runs the same.


package.json
```diff
@@ -41,22 +42,18 @@
   ],
   "license": "MIT",
   "scripts": {
-    "build": "npm run clean && npm run build-alt && npm run build-utils && npm run build-alt-browser && npm run build-alt-browser-with-addons",
-    "build-alt": "babel src/alt --out-dir lib --stage 0",
```
goatslacker: hmm let me play with this because I like having this at the moment so I can just selectively build alt and not have to wait until util builds

package.json
```diff
@@ -27,6 +27,7 @@
     "jsdom": "^3.1.2",
     "lunr": "^0.5.9",
     "mocha": "^2.2.4",
+    "object-assign": "^2.0.0",
```
goatslacker: ??

.gitignore
```diff
@@ -1,5 +1,6 @@
 TODO
 coverage
+dist
```
goatslacker: i definitely don't want to ignore this

package.json
```diff
@@ -41,22 +42,18 @@
...
-    "posttest": "npm run build-alt",
     "prepublish": "npm run lint && npm run test && npm run build",
-    "pretest": "npm run clean && npm run build-test",
-    "size": "npm run build-alt; browserify flux.js > flux-build.js; uglifyjs -m -c 'comparisons=false,keep_fargs=true,unsafe=true,unsafe_comps=true,warnings=false' flux-build.js  > flux-build.min.js",
-    "test": "npm run tests-node",
+    "pretest": "npm run clean && npm run transpile",
+    "test": "npm run test-node",
     "test-browser": "browserify test/browser/index.js -t babelify --outfile test/browser/tests.js",
-    "tests-node": "mocha -u exports -R nyan --require ./test/babel test"
+    "test-node": "mocha -u exports -R nyan --require ./test/babel test"
```
goatslacker: this looks cleaner though! :sunglasses:

package.json
mikekidder: needed for -> "test-browser": "browserify test/browser/index.js -t babelify --outfile test/browser/tests.js"

.gitignore
mikekidder: yeah, some repos dont include dist but see why yours does. I can live with the git status changes when it builds

package.json
goatslacker: oh right, :+1:

.gitignore
goatslacker: yeah it's annoying, wish there was a better way to provide the browser builds, maybe I should post them somewhere instead of checking them in.

goatslacker:
Thanks!

PR #284, opened 2015-06-01
The holy-grail of server rendering

goatslacker:
There's some clean up work I'd still like to do with this but I'm throwing it up here for now to gather early feedback/let people know what's coming.

Why?

It allows us to define our data fetching needs in the component. It's not quite relay though, you don't define what you need but rather how you're going to get it.

This works very well with Alt's [Data Sources](http://alt.js.org/docs/async/).

This lets you keep your singleton stores and actions which are easier to reason about and you don't have to pass flux around in context.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "env": {
     "node": true,
+    "browser": true,
     "es6": true
   },
   "ecmaFeatures": {
--- a/dist/alt-with-addons.js
+++ b/dist/alt-with-addons.js
@@ -1568,18 +1568,24 @@ var StoreMixin = {
           return x;
         };
 
+        var makeActionHandler = function makeActionHandler(action) {
+          return function (x) {
+            var fire = function fire() {
+              loadCounter -= 1;
+              action(intercept(x, action, args));
+            };
+            return typeof window === 'undefined' ? function () {
+              return fire();
+            } : fire();
+          };
+        };
+
         // if we don't have it in cache then fetch it
         if (shouldFetch) {
           loadCounter += 1;
           /* istanbul ignore else */
           if (spec.loading) spec.loading(intercept(null, spec.loading, args));
-          spec.remote.apply(spec, [state].concat(args)).then(function (v) {
-            loadCounter -= 1;
-            spec.success(intercept(v, spec.success, args));
-          })['catch'](function (v) {
-            loadCounter -= 1;
-            spec.error(intercept(v, spec.error, args));
-          });
+          return spec.remote.apply(spec, [state].concat(args)).then(makeActionHandler(spec.success))['catch'](makeActionHandler(spec.error));
         } else {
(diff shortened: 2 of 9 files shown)


src/utils/forRender.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const withData = Render.withData
+
+export default function forRender(alt, Component, props) {
+  // recycle state
+  alt.recycle()
+
+  /*eslint-disable */
+  if (typeof window === 'undefined') {
+    /*eslint-enable */
```
iam4x: Since you added `browser: true` in eslint configuration shouldn't you remove the enable/disable comments?

goatslacker: yeah. I'm going to either remove browser: true or remove these.

I just threw browser: true in there so lint would shut up so I could push this.

transedward:
> This lets you keep your singleton stores and actions which are easier to reason about and you don't have to pass flux around in context.

Would you like to elaborate on how this work inside?

srph:
Interesting. I tend to put *request*s on the components itself, and just *update* / *hydrate* the *store*s with the *response*.

goatslacker:
@transedward sure.

At a high level:

* All the stores are flushed from any previous data.
* We resolve all the data fetches that need to be made in a single pass and then execute them.
* We trap all the dispatches so they buffer the calls.
* Once we have all the data back in a Promise.all we unleash all the dispatches in order. Dispatches are synchronous.
* Once all dispatches have completed we render the component tree again which should have the up to date data from the stores.
* We send that markup down through the promise so one can render it server-side.

transedward:
@goatslacker thanks for such detailed step by step explanation
do we have guarantee that each request can be handle concurrently even if without multiple instances of flux?
Is that because you mention
> We trap all the dispatches so they buffer the calls.
so that program won't be blocked until Promise.all resolved and then all the dispatches dispatched?

goatslacker:
If the tests are passing then we have that guarantee: https://github.com/goatslacker/alt/blob/master/test/alt-iso-test.js#L124

it is because we trap dispatches, it is non-blocking however, concurrent requests are allowed.

iam4x:
@goatslacker I have a few questions about `AltIso.render`

* Is it possible to resolve promise from a child component? For instance I have an App component that will resolve internationalised messages and which render an User component that will resolve an array of users too. I will see only the first component doing the fetch (App for messages).
* Have you tried to run it with `react-router`? After two transitions/page change it won't render the new component.

I'm trying to figure out but I'm getting stuck now :/ Help would be highly appreciated for updating the boilerplate (https://github.com/iam4x/isomorphic-flux-boilerplate/pull/69)

iam4x:
For `react-router` on the first transition, when `AltIso.render` is called: `toDOM` will unmount the whole component it will call `HistoryLocation.removeChangeListener` from `react-router` and will not be able to render a again the app.

PR #287, opened 2015-06-02
Using Alt.js in Maple example

Wildhoney:
[Maple](https://github.com/Wildhoney/Maple.js) uses Alt.js in its example for webcomponents in React &ndash; with an [example up and running on Heroku](http://maple-app.herokuapp.com/).


jdlehman:
Awesome, thanks for sharing!

PR #291, opened 2015-06-02
Reverse then/catch in DataSources

### LINKED ISSUE #290: Reverse then/catch in data sources
We don't want to catch errors in the store's success handler...

jfsiii:
Catch errors thrown by the fetch; not the success handler. fixes #290

### DIFF
--- a/src/alt/store/StoreMixin.js
+++ b/src/alt/store/StoreMixin.js
@@ -67,8 +67,8 @@ const StoreMixin = {
           /* istanbul ignore else */
           if (spec.loading) spec.loading(intercept(null, spec.loading, args))
           return spec.remote(state, ...args)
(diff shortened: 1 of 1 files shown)


goatslacker:
:+1: 

I tried this yesterday but this fails the tests currently, I have to look into why that is.

jfsiii:
Ah, sorry about that. I should have run the tests locally first.

goatslacker:
Got it, it's because the catch should throw otherwise the catch will run and then the 'then' will run too!

This decrements our loading counter and is fucking up our tests, but most important it makes the behavior of data sources weird since fail and success will be called.

The solution is to throw in the catch handler.

goatslacker:
8e1418b854bd59f3d1a0e99e353cf997bb52bdf0

got it

PR #295, opened 2015-06-03
Return from action to dispatch

goatslacker:
First bits I'm stealing off of @gaearon returnable actions which dispatch.


goatslacker:
Was merged in 460a749c28c0156bbf8bae5e2c791cf3a07f3c42

PR #299, opened 2015-06-03
Add some notes in the documentation about how to create public store methods using es5 syntax. 

david0178418:
I had to dig around to figure out that a "publicMethods" property could be defined when using the object syntax.  I added this to the es5 guide for anyone else looking for this.  While I was at it, I added examples of the other methods as well.


goatslacker:
Awesome. I love docs PRs! Thanks

david0178418:
Glad to help! Thanks for the hard work on this.

PR #302, opened 2015-06-04
Update Bower package to 0.16.9

mattkahl:
Stop leaving us lowly Bower users out in the cold after the version bump. ;)

Changes:
- Bower will now serve 0.16.9 as the `bower install` default

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "alt",
-  "version": "0.16.6",
+  "version": "0.16.9",
   "homepage": "https://github.com/goatslacker/alt",
   "authors": [
     "Josh Perez <josh@goatslacker.com>"


goatslacker:
I wish bower would just ditch the bower.json crap and go with package.json so `npm version patch` would just bump both. If you have any release tips I'd love to hear them!

Thanks for patch.

mattkahl:
Too many standards; [let's create another one](https://xkcd.com/927/)!

I don't have an opinion based on experience, but it looks like both of these are viable options (in descending order of popularity):
- https://www.npmjs.com/package/mversion
- https://www.npmjs.com/package/versiony

With each of those, you're able to update the version of all/any relevant packaging files in one fell swoop.

e.g. (From the docs)
- CLI Usage:
```
$ mversion 0.0.5 -m
New Version: 0.0.5
Updated package.json
Updated component.json
Updated manifest.json
Commited to git and created tag v0.0.5
```
- Node usage
```
mversion.update({
    version: '0.0.1',
    commitMessage: 'Bumping version'
  }, function (err, data) { })
```

PR #306, opened 2015-06-06
Component did connect hook

ericwooley:
I added an API for component did connect, I am not sure if this was where you wanted it in the source, not quite sure how to ensure it makes it into the decorator. Let me know if you want me to change it.

### DIFF
--- a/src/utils/connectToStores.js
+++ b/src/utils/connectToStores.js
@@ -66,8 +66,8 @@ function connectToStores(Component) {
       stores.forEach((store) => {
         store.listen(this.onChange)
       })
-      if (typeof Component.componentDidConnect === 'function') {
-        Component.componentDidConnect()
+      if (Component.componentDidConnect) {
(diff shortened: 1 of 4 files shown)


goatslacker:
I was thinking it would be a static function similar to how getPropsFromStores is already

ericwooley:
If you make it a static, then it will fire every time any of the components are mounted, and you would have no access to this. 

Seemed weird to me that it would fire for every mount, but not have access to this. I can switch it if you want though, then it wouldn't require a ref, which is nice.

src/utils/connectToStores.js
```diff
@@ -66,6 +66,9 @@ function connectToStores(Component) {
       stores.forEach((store) => {
         store.listen(this.onChange)
       })
+      if (typeof Component.componentDidConnect === 'function') {
```
goatslacker: we can assume its a function

and we should pass in props

PR #318, opened 2015-06-10
Update README with React Native example

mrblueblue:
I've added a React Native example application to the examples list.

The application renders the most recent tweets for a given location and was built using React Native with an Alt Flux implementation. 

You can see the repo [here](https://github.com/mrblueblue/react-native-alt-demo)


goatslacker:
Amazing! Thanks

Issue #319, opened 2015-06-10
Change is emitted after Store's action handler even when state is not changed


endragor:
It looks like alt code emits change event from Stores even when nothing was actually changed, i.e. when `setState` was not even called in the handler. I wonder what's the reason behind such behaviour, if it's intended (it seems that it is, looking at the code).
I have cases when `setState` might not be called from the action handler under certain circumstances. Besides, `setState` might be called to set a primitive value (boolean, number or string) that already has the same value (for example, boolean flag was already set to `true`). To prevent this, I have to manually insert `this.preventDefault()` and manually check whether my `setState()` is actually going to change something to save performance. I would love if the framework did it for me. I know it might be not reasonable to always compare the whole new state and old state, but just comparing the primitive values would already cover many cases, in my opinion.

mercuriete:
whenever you return something different to false then the store emit a change.

i had this problem before.
if you debug the handler and make some steps after the return you will see that alt emit a change if you return something diferent than false.

anyway, remember to use immutable js for your stores in order to know quicky is something has changed with only a single reference comparison.

I hope this was helpful. :)

endragor:
@mercuriete 
Thanks for your comment.

I know that a change is always emitted, unless you return `false`. And that's what my concern is about. I don't want a change to be emitted, unless I explicitly call `setState`

I don't understand the rationale behind the current behaviour.

jdlehman:
I believe the rationale is that alt is not forcing people to use `setState`, though it is recommended.  So someone can just set data directly on the store `this.myData = 1`, which is why emitting changes is the default behavior.  In this case it makes sense because then it does not require the user to explicitly emit the change, though the tradeoff is that they have to explicitly return false if they do not want to emit a change.  `setState` actually was a more recent addition to alt, but we did not change the default behavior for this (and if we did it would be a breaking change and not show up until a major release).

jdlehman:
Also you can use [`preventDefault`](https://github.com/goatslacker/alt/blob/master/docs/createStore.md#storemodelpreventdefault) in the store to do the same thing as returning false:

endragor:
I see. Thanks for the explanation. Are there any plans to make alt configurable to be able to change the behaviour like the one being discussed? We never change Store fields directly, so we don't need the current behaviour.

jdlehman:
I don't see why not, we generally like providing the ability to configure.

[...]

PR #324, opened 2015-06-11
Add shouldStoreUpdate config to Store

jdlehman:
- Provides config to set whether emitChange should be called automatically or not after store action listeners

### DIFF
--- a/src/alt/store/AltStore.js
+++ b/src/alt/store/AltStore.js
@@ -49,13 +49,14 @@ class AltStore {
 
       const actionHandler = model.actionListeners[payload.action] ||
         model.otherwise
+      const shouldEmitChange = this.StoreModel.config.shouldEmitChange
 
       if (actionHandler) {
         const result = handleDispatch(() => {
           return actionHandler.call(model, payload.data, payload.action)
         }, payload)
 
-        if (result !== false && !this.preventDefault) this.emitChange()
+        if (result !== false && !this.preventDefault && shouldEmitChange) this.emitChange()
       }
 
       if (model.reduce) {
@@ -63,7 +64,7 @@ class AltStore {
           model.setState(model.reduce(this.state, payload))
         }, payload)
 
-        if (!this.preventDefault) this.emitChange()
+        if (!this.preventDefault && shouldEmitChange) this.emitChange()
       }
 
       this.lifecycle('afterEach', {
--- a/src/alt/store/AltStore.js
+++ b/src/alt/store/AltStore.js
@@ -56,7 +56,9 @@ class AltStore {
           return actionHandler.call(model, payload.data, payload.action)
         }, payload)
 
-        if (result !== false && !this.preventDefault && shouldEmitChange) this.emitChange()
+        if (result !== false && !this.preventDefault && shouldEmitChange()) {
+          this.emitChange()
+        }
       }
 
       if (model.reduce) {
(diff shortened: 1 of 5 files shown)


jdlehman:
This is a starting point for Issue #319. Still needs tests before merging.

I wasn't sure if we wanted `shouldStoreUpdate` to work in conjunction with returning `false` or using `preventDefault` or if it should have precedence in determining whether `emitChange` gets called.

src/alt/store/AltStore.js
```diff
@@ -55,7 +55,8 @@ class AltStore {
           return actionHandler.call(model, payload.data, payload.action)
         }, payload)
 
-        if (result !== false && !this.preventDefault) this.emitChange()
+        const shouldUpdate = this.StoreModel.config.shouldStoreUpdate && this.StoreModel.config.shouldStoreUpdate(this.state)
```
endragor: How can you compare old state with new state here?
In React you can use `this.state` and `this.props` from `shouldComponentUpdate` to get the current state/props, and compare them with the new state/props, passed as arguments.

jdlehman:
Optionally we could just add a config property instead of a function that specifies whether store action listeners emit automatically or not.

endragor:
Yes, that would satisfy my requirement.

goatslacker:
:+1: to that idea

src/alt/store/AltStore.js
```diff
@@ -55,7 +55,7 @@ class AltStore {
           return actionHandler.call(model, payload.data, payload.action)
         }, payload)
 
-        if (result !== false && !this.preventDefault) this.emitChange()
+        if (result !== false && !this.preventDefault && this.StoreModel.config.shouldEmitChange) this.emitChange()
```
goatslacker: nit: can we pull out shouldEmitChange somewhere so this minifies better?

goatslacker:
Food for thought: there's an output() method that exists. The purpose of this method is to transform output that is emitted, perhaps if we return undefined from this method then it can suppress the change?

src/alt/store/AltStore.js
endragor: If `shouldEmitChange` is false, how will emitChange() ever be called? It should still be called if `setState` was called at least once.

Since `shouldEmitChange` is not a function anymore, a better name might be `alwaysEmitChange`.

jdlehman: @endragor I believe `emitChange` will still be called when `setState` is used [here](https://github.com/goatslacker/alt/blob/master/src/alt/store/index.js#L24). @goatslacker correct me if I am wrong.

jdlehman:
As a note, I will be out of pocket for a week (vacation), so feel free to move this forward in my absence.

goatslacker:
Will do, just missing some tests and making it a function call.

cesarandreu:
This is useful if you're using immutable.js, since you might have multiple stores listening to an action but that doesn't mean that they all need to update.

There's a similar fluxible util: [createImmutableStore](https://github.com/yahoo/fluxible-immutable-utils/blob/master/lib/createImmutableStore.js). They do a [state.equals(nextState) check](https://github.com/yahoo/fluxible-immutable-utils/blob/master/lib/createImmutableStore.js#L24-L34) that would be nice to add to alt's ImmutableUtil.

goatslacker:
This is only useful if we make it so that:

shouldStoreUpdate(nextState, currentState)

PR #325, opened 2015-06-11
Update the docs and guide to use ES6 modules

nikhilbaradwaj:
Update the docs and guide to use ES6 modules


goatslacker:
This was actually on purpose so people would follow the guide a bit easier without having to digest a bulk of ES6 along with flux and alt.

nikhilbaradwaj:
I see. OK, thanks. I'll close the request.

PR #327, opened 2015-06-11
fix todo item text update bug (examples/todomvc)

ericsong:
https://github.com/goatslacker/alt/issues/326


examples/todomvc/js/stores/TodoStore.js
```diff
@@ -37,7 +37,9 @@ var todoStore = alt.createStore(class TodoStore {
   }
 
   onUpdateText(x) {
-    var { id, text } = x
+    var id = x.id,
```
goatslacker: We can leave this line as is

ericsong:
got it. reverted the change

PR #340, opened 2015-06-19
Hoist non-react statics when using connectToStores

cesarandreu:
I noticed that alt only has flux and transmitter as its dependencies... So I'm not sure if you'd wanna add `hoist-non-react-statics` just for connectToStores, which admittedly many people probably won't be using. 

This package was made for fluxible's connectToStores, and I think it'd be useful for alt's connectToStores  to share this behavior.

If you'd prefer to avoid adding the dependency I could add the hoist-non-react-statics code to connectToStores, or another alternative is for me to publish the tweaked connectToStores as a separate module. 

It's also worth noting that it only works with `React.createClass`, since `class` statics are not enumerable. It won't break with class syntax, it just won't hoist statics.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "lib",
   "dependencies": {
     "flux": "2.0.3",
+    "hoist-non-react-statics": "^1.0.2",
     "transmitter": "1.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


goatslacker:
You'd probably want to use something like: https://github.com/goatslacker/alt/blob/master/src/utils/statics.js

Example:

```js
statics({
  getFoo() {
    return 'foo'
  },
}, connectToStores({
  getPropsFromStores() { },
  getStores() { },
}, MyComponent))
```

The example is organized a bit ugly but you can see the gist of it.

PR #347, opened 2015-06-23
my contribution

darul75:
Have tried implement some of great Alt best practices...work is still in progress when I got time :) but snapshots...server side rendering, hot reload, good stuffs there ;)


goatslacker:
Sure thing!

PR #353, opened 2015-06-25
Pass source args to loading, success, and error actions

cesarandreu:
This PR makes it so that source args are passed to loading, success, and error actions. 

* Not sure if it should spread or pass it as a single argument
* Open to suggestions for how to make the tests better

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -89,7 +89,7 @@ This function is called first. If a value is returned then a change event will b
 
 This function is called whenever we need to fetch a value remotely. `remote` is only called if `local` returns null or undefined as its value, or if `shouldFetch` returns true.
 
-Any arguments passed to your public method will be passed through to both local and remote:
+Any arguments passed to your public method will be passed through to both local and remote, as well as to the loading, success, and error actions:
 
 ```js
 remote(state, one, two, three) {
@@ -98,15 +98,19 @@ remote(state, one, two, three) {
 SearchStore.performSearch(1, 2, 3);
 ```
 
-### success
+### success(resolveValue: any, ...args: any)
 ##### Required
 
 Must be an action. Called whenever a value resolves.
 
-### error
+Called with any arguments passed to the public method.
+
+### error(rejectValue: any, ...args: any)
 ##### Required
 
 Must be an action. Called whenever a value rejects.
+
+Called with any arguments passed to the public method.
 
 ### shouldFetch(state: object, ...args: any)
 _Optional_
@@ -124,10 +128,14 @@ interceptResponse(data, action, args) {
 }
 ```
 
-### loading
+### loading(value: any, ...args: any)
 _Optional_
 
 Must be an action. This function will be called immediately prior to the `remote` function. This call is a synchronous  and is not related to the `isLoading()` method of the store.
+
+Value will be null unless changed by interceptResponse.
+
+Called with any arguments passed to the public method.
 
 ## Decorator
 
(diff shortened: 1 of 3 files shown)


jdlehman:
Good stuff, I prefer the spread over a single argument, but that is just my preference.  We will also want to make sure the docs get updated before this goes out.

I am wondering if we want to allow for different arguments to be passed to each action (loading, success, and error).  With this you could still accomplish it, but you might have to pass arguments to success that you only need for error or vice versa.

cesarandreu:
Great! I prefer spread as well, and it was my initial choice. Happy I don't have to change it.

I added some docs, open to suggestions for how to improve em.

With respect to what you mentioned, I figured if you want something more fine-grained, you'd already use interceptor for that. The reason I want to include args is so that I can do something such as `FooStore.destroyFoo({ foo_id: 123})`, and when it succeeds I can clear `foo_id` in my store.

jdlehman:
@goatslacker thoughts?

goatslacker:
Oh we hashed it out on reactiflux. This is :+1: pending some docs.

cesarandreu:
I updated `docs/async.md`, is there anywhere else that needs updating?

jdlehman:
Docs look good to me.  We need to get our Travis builds in order... it's been a problem lately.

goatslacker:
Babel needs to be updated.

This branch needs to be rebased though.

goatslacker:
Ok I updated babel, this just needs a rebase and we're :+1:

cesarandreu:
I just rebased and updated the docs, take a look?

goatslacker:
It's good. Lets wait on Travis.

goatslacker:
Looks like there are failing tests. Let me look into this.

goatslacker:
Ugh. I'll fix this one...

cesarandreu:
When I run it locally I get 4 errors :(. 

```
  1)  the actions get the alt instance:
     AssertionError: alt is an instance of Alt: expected undefined to be an instance of Alt
      at ActionsGenerator.MyActions (test/stores-get-alt.js:20:16)
      at new ActionsGenerator (lib/index.js:158:92)
      at lib/index.js:183:30
      at Alt.createActions (lib/index.js:184:10)
      at Context.theActionsGetTheAltInstance (test/stores-get-alt.js:24:9)

  2) AltIso browser browser requests:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  3) AltIso concurrent server requests are resolved properly:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.


  4) decorators decorating actions with args:
     AssertionError: expected undefined to be a function
      at Context.decoratingActionsWithArgs (test/decorators-test.js:58:14)
```

goatslacker:
If you rebase and npm install you'll only get 1. Maybe 2. I fixed one bug last night will look into why this one is failing today.

cesarandreu:
Rebased off master, let's see how it goes this time :D.

goatslacker:
I know what's wrong...

so alt.generateActions will pass your dispatches through as an array which means that passing it in as more than one arg will result in broken behavior.

Example:

```js
const foo = alt.generateActions('bar')

foo.bar(1, 2, 3)

// when listening to foo.bar in a store:

bar([one, two, three]) {
}


// why is this a problem?
// so you used to be able to do this:

const foo = alt.generateActions('bar')
source.success = foo.bar

store.bar(data) {
  // handle data here.
}

// but now data is an array :(
```

Issue #350, opened 2015-06-25
Dispatch doesn't seem to complete until listener finishes?


bryannaegele:
TL;DR
Calling a transition on react router in the store listener of a view throws a "Cannot dispatch in the middle of a dispatch" error when the route you are transitioning to has a view which invokes a source fetch on another store in componentDidMount.

I'm seeing a problem when a component mounts (with a store fetch call in componentDidMount) raising a "cannot dispatch in the middle of a dispatch" error.

Basic setup is my header component is listening to a user store. Logging in calls a user login action which dispatches success/failure. The header's onUserChange handler I have for the store then looks at the incoming state. If the user is now authenticated, it calls a transition on react router to the user's home page. The page being transitioned to has a component that does a fetch on another store in componentDidMount.  This throws an error about being in the middle of a dispatch.

```javascript
export default class HeaderMenu extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      user: UserStore.getState().user
    };
  }
  static getProps() {
    return {};
  }
  componentDidMount = () => {
    UserStore.listen(this.onUserChange);
  }

  componentWillUnmount = () => {
    UserStore.unlisten(this.onUserChange);
  }

  onUserChange = (nextState) => {
    let user = nextState.user;
    let currentState = this.state.user;

    // not currently authenticated, becomes authenticated
    if(currentState.get('authenticated') === false && user.get('authenticated')) {
      this.setState({
        user: user
      });
      // invoking these transitions here or in the callback of the setState does not make a difference
      this.context.router.transitionTo('myProjects', { organization: user.get('realm')});
    } else if (currentState.get('authenticated') && user.get('authenticated') === false) {
      // currently authenticated, becomes unauthenticated
      this.setState({
        user: user
      });
      this.context.router.transitionTo('login');
    } else if( !Immutable.is(currentState, user)) {
      this.setState({
        user: user
      });
    }
  }

  render() {
    var {router} = this.context;
    let user = this.state.user;
    let classes = cn('headerMenu', styles.this);
    if (user.get('authenticated')) {
      return <div className={classes}>
        <LoggedInMenu user={user} />
      </div>;
    } else {
      return <div className={classes}>
        <LoggedOutMenu />
      </div>;
    }
  }
}

HeaderMenu.contextTypes = {
  router: React.PropTypes.func.isRequired
};
```
And the page transitioning to...

```javascript
export default class MyProjectsPage extends React.Component {
  constructor(props) {
    super(props);
[...]

PR #355, opened 2015-06-26
Added more clarity for implementing sources in async doc

bryannaegele:
Added some possibly helpful pieces to the async doc for sources around required/optional methods, loading method, and ways to always fetch remotely.

Closes goatslacker/alt#350


goatslacker:
Nice. Thanks!

Issue #361, opened 2015-06-27
underdash: Stores/Data Fetching


taion:
ref #337 

I'm using this one issue for both concerns because both Marty and Alt encourage a pattern of handling async data loads from stores:
- http://alt.js.org/docs/async/
- http://martyjs.org/guides/fetching-state/index.html

Stores seem pretty similar otherwise - set up handlers with respectively `@bind` and `@handles`, &c. Alt hides store methods by default. Marty doesn't. Maybe more intuitive if handlers are hidden by default but other methods are exposed?

Marty has a concept of state sources to allow changing data fetch (and modification) behavior when running on the server - these should probably be factored out into a separate library.

taion:
One other difference - Alt data sources are like actions that load from remote if necessary. Marty store fetches just give the appropriate value back to the container, or indicate pending/failed state to the container if necessary, and explicitly handle things like having only a single pending fetch per fetch ID.

I think a combined implementation that merges the best from both approaches might be:

```js
class UserStore extends Store {
  getUser(id) {
    return this.fetch({
      id: id,
      local: () => this.state[id],
      remote: () => fetch(`/api/users/${id}`).then(res => res.json()),
      loading: UserActions.getUserStarting,
      success: UserActions.getUserDone,
      failed: UserActions.getUserFailed
    });
  }
}
```

Instead of a custom fetch result, this could just return the value, or special sentinel `LOADING` or `FAILED` values that containers can interpret.

A couple of extensions might be:

* Accept synchronous callbacks instead of actions for the `loading`/`success`/`failed` handlers (to allow e.g. dispatching multiple actions when used with something like https://github.com/gaearon/normalizr)
* Allow specifying custom `pendingValue` and `failedValue` for more granular component-level control of pending/failed handling (rather than at the container level)

taion:
One more implementation-level difference - Marty stores explicitly keep track of pending/failed fetches: https://github.com/martyjs/marty-lib/blob/v0.10.6/src/store/store.js#L26-L28. These (probably) don't need to be preserved when dehydrating, but render loops would need to preserve these values.

taion:
Copying over my proposal from #364:

-----

https://github.com/goatslacker/alt/pull/364#issuecomment-116405580

I agree that, right now, Alt fetch sometimes does not benefit from specifying the `local` branch. However, I feel that this is something of a missed opportunity. You still need to specify where the data comes from on the store to render it in the component, so you end up potentially repeating yourself.

[...]

PR #364, opened 2015-06-28
A sweet spankin' new Render

goatslacker:
* Changes Render API to be more sensible?
* Performs a few iterations when rendering so we can render child components that need data resolved.
* Partial rendering, if one component errors it doesn't taken down the entire render tree. We'll render what we've got and then client side render the rest.
* No need to perform if-checks in your React render methods, all components are only rendered if their props are resolved. We keep track of what's been resolved and don't re-fetch client side.
* You can still keep your singletons and have concurrent server requests.
* Now you get a spiffy error object telling you what went wrong. No more need to .then().catch() it, everything is just returned in .then()
* You also get a nice diagnostics object telling you how many iterations it took, and how long it took (in ms) to resolve + render.

@taion @jfsiii @jdlehman

### DIFF
--- a/src/utils/Render.js
+++ b/src/utils/Render.js
@@ -4,27 +4,36 @@ export function withData(fetch, MaybeComponent) {
   function bind(Component) {
     return React.createClass({
       contextTypes: {
+        universalId: React.PropTypes.string.isRequired,
         buffer: React.PropTypes.object.isRequired
       },
 
       childContextTypes: {
+        universalId: React.PropTypes.string.isRequired,
         buffer: React.PropTypes.object.isRequired
       },
 
       getChildContext() {
-        return { buffer: this.context.buffer }
+        const children = this.props.children || []
+        const universalId = `${this.context.universalId}.${children.length}`
+        return {
+          universalId,
+          buffer: this.context.buffer
+        }
       },
 
       componentWillMount() {
-        if (!this.context.buffer.locked) {
+        if (this.context.buffer.shouldFetch(this.context.universalId)) {
           this.context.buffer.push(
+            this.context.universalId,
             fetch(this.props)
           )
         }
       },
 
       render() {
-        return this.context.buffer.locked
+        // XXX I need to make sure I can render since this promise has fulfilled successfully!
+        return this.context.buffer.shouldRender(this.context.universalId)
           ? React.createElement(Component, this.props)
           : null
       }
@@ -42,11 +51,15 @@ function call(f) {
 function usingDispatchBuffer(buffer, Component) {
   return React.createClass({
     childContextTypes: {
+      universalId: React.PropTypes.string.isRequired,
       buffer: React.PropTypes.object.isRequired
     },
 
     getChildContext() {
-      return { buffer }
+      return {
+        universalId: 'root',
+        buffer
+      }
     },
 
     render() {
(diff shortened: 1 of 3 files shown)


taion:
One thought - right now you're using context to pass around the dispatch buffer. However, if you restrict data fetches to data source operations from stores, don't the stores already have a reference to the application? In which case you might be able to get around the use of React's context altogether, and avoid the random issues that pop up there.

taion:
Wait, I see that the async methods already checks for buffer on `alt`. Why not just append to a buffer array (or something) there?

goatslacker:
In which case the application will have to be an instance so that requests don't collide, so we'd still have to use context to pass around the application.

We can't use the buffer in alt because alt is a singleton, so async requests will be butchering the alt buffer.

I don't mind using context under the hood as long as as a developer I don't have to deal with it.

taion:
`React.render` is synchronous, no? So you could do at https://github.com/goatslacker/alt/pull/364/files#diff-4bd8b153d7a5a02c9d41afdd3e91c113R80

```js
alt.buffer = [];

const html = this.renderStrategy(...);

const promisesBuffer = alt.buffer;
alt.buffer = null;
```

or is there something here I'm missing?

src/utils/Render.js
```diff
@@ -115,23 +135,110 @@ function renderWithStrategy(strategy) {
...
+    if (opts.fulfilled) buffer.fulfilled = opts.fulfilled
+    const Node = usingDispatchBuffer(buffer, Component)
+    const Element = React.createElement(Node, props)
+    buffer.clear()
+    return React.render(Element, documentNode)
+  }
+
+  static resolve(fetch, MaybeComponent) {
+    function bind(Component) {
+      return React.createClass({
```
taion: Might be preferable to modify the component prototype in-place. I know it's a bit messy but otherwise if you export a component that's been e.g. decorated with `@resolve`, the exported component loses all the statics and public getter methods and anything else you might care about. This has bitten me in the past with containers, but it's unavoidable there, whereas it is avoidable here.

goatslacker: Not a bad idea...

Although we do have https://github.com/goatslacker/alt/blob/master/src/utils/statics.js for that sort of thing.

I'll amend the changes.

goatslacker:
What about the rest of the stuff? I need to keep a list of what's been been fulfilled/fetched so I don't refetch. I can keep the list in a closure but the components need a way to query that.

I also need to store an id for each component, although I think I can do that without context.

taion:
Don't you just hit the `local` branch on the data source for fulfilled fetches?

taion:
ref https://github.com/goatslacker/alt/issues/361

I don't know if this is in-scope for this change, but a couple of things on the Marty fetch that can help with this:

1. Associate an ID with each "fetch" at the store level, and only allow a single pending remote fetch per ID: https://github.com/martyjs/marty-lib/blob/v0.10.6/src/store/storeFetch.js#L50-L52
2. After the remote fetch completes, check that the local result is present - if not, consider this a failure and insert an appropriate "failed" sentinel

taion:
Thinking about this a little more, if we're targeting this for the underdash milestone, I'd suggest that we resolve #361 first.

* If we move the bookkeeping for fetches onto the store, you end up with an API that's more user-friendly because it will collapse pending requests on the same ID for you, and you don't need the extra bookkeeping for `Render`
* If we make these fetch calls return the actual value (potentially with loading/failed/not found sentinels), then we can remove the `define` concept and just have containers handle everything, and reduce API surface area

goatslacker:
Relying on hitting the local branch puts the burden on the developer to figure out if its been fetched/fulfilled or not.

I'm currently associating an ID per component since components export only a single fetch via Promise.all. It's an all or nothing approach since components need all the props to resolve in order to render.

Oh I also can't modify the component in place, at least I don't think I can. I need to control render so I can suppress render until all props resolve (similar to Relay)

What extra book-keeping are we hoping to avoid by moving the fetches to the store?

I was aiming at shipping this with 0.17.0 (since 0.16.10 is currently broken) and then building on it for the malt/underdash.

taion:
>Relying on hitting the local branch puts the burden on the developer to figure out if its been fetched/fulfilled or not.

Isn't that the intended design though? i.e. outside of the render loop context, check for the data being present locally before fetching remotely.

>I'm currently associating an ID per component since components export only a single fetch via Promise.all. It's an all or nothing approach since components need all the props to resolve in order to render.
>
>What extra book-keeping are we hoping to avoid by moving the fetches to the store?

If the store keeps track of pending/failed loads, then render doesn't need to keep track of this, and (if individual loads are keyed by ID), you don't end up with two requests if two components request the same data at the same time.

>Oh I also can't modify the component in place, at least I don't think I can. I need to control render so I can suppress render until all props resolve (similar to Relay)

Oh you're right. Short of picking up `render` from the prototype and wrapping it... and that'd be a bit silly.

>I was aiming at shipping this with 0.17.0 (since 0.16.10 is currently broken) and then building on it for the malt/underdash.

This makes sense then. I think I'd like to merge `define` with a container that can render an empty `div` or something while loading post underdash-ing, but no point in making multiple sets of breaking changes when not needed.

goatslacker:
I want the least possible developer involvement. Defining local yourself can be great if you want to optimize, but in an ideal world you don't have to think about it and the library will just do the right thing. Sometimes I like to forego local because on the client side I always want to fetch, but remotely I would only like to fetch once pre-render. Having to perform that logic myself can/will suck.

taion:
Apologies, I've been conflating a couple of topics here.

I agree that, right now, Alt fetch sometimes does not benefit from specifying the `local` branch. However, I feel that this is something of a missed opportunity. You still need to specify where the data comes from on the store to render it in the component, so you end up potentially repeating yourself.

If I'm not mistaken, right now you have something like (apologies in advance for mangling Alt syntax)

```js
const UserSource = {
  fetchUsers: {
    remote: () => UserApi.getUsers,
    loading: UserActions.getUsersStarting,
    success: UserActions.getUsersDone
    error: UserActions.getUsersFailed
  }
}

@datasource(UserSource)
class UserStore {
  @bind(UserActions.getUsersDone)
  receiveUsers(users) {
    this.users = users;
  }
}
const userStore = alt.createStore(UserStore);

@AltIso.define(props => UserStore.fetchUsers())
class BoundUserList extends React.Component {
  render() {
    return (
      <AltContainer 
        store={userStore}
      >
        <UserList />
      </AltContainer>
    );
  }
}

class UserList extends React.Component {
  render() {
    return (
      <div>{JSON.stringify(this.props.users)}</div>
    )
  }
}
```

I think the way the stores, the fetch operation, and the containers should interact should look more like e.g.:

```js
@alt.register
class UserStore extends Store {
  getUsers() {
    return this.fetch({
      local: () => this.users,
      remote: () => UserApi.getUsers(),
      loading: UserActions.getUsersStarting,
      success: UserActions.getUsersDone
      error: UserActions.getUsersFailed,
      refresh: true
    });
  }

  @bind(UserActions.getUsersDone)
  receiveUsers(users) {
    this.users = users;
  }
}

class UserList extends React.Component {
  render() {
    return (
      <div>{JSON.stringify(this.props.users)}</div>
    )
  }
}

const BoundUserList = createContainer(
  UserList,
  {
    store: userStore,
    props: {
      users: () => userStore.getUsers()
    }
  }
);
```

The idea is - you combine the get and fetch operations, and you have the container be smart enough to not render its contained child if there are pending values.

-----

The other part is, by using stores to keep track of things, I mean putting those things to be kept track of (pending and failed fetches) as part of the state that (if I'm not misunderstanding) is getting flushed and bootstrapped between iterations of the render loop.

As long as you take care of bootstrapping the stores before you call `render`, then flushing them afterward, the singleton-ness of the stores shouldn't be an issue, no?

taion:
To be clear, what I mentioned above is something like what I'd look for w/the underdashed project - essentially a sketch of an API that combines Marty fetch/containers with Alt idioms that should work with the render loop. And this API is better for cases where you don't want to re-fetch the data; when you do, you end up writing something like

```js
const BoundUserList = createContainer(
  UserList,
  {
    store: userStore,
    props: {
      users: () => userStore.getUsers()
    },

    componentWillMount() {
      userStore.refreshUsers();
    }
  }
);
```

which is a little nasty. Maybe instead something like:

```js
const BoundUserList = createContainer(
  UserList,
  {
    store: userStore,
    props: {
      users: {
        value: () => userStore.getUsers(),
        refresh: true
      }
    }
  }
);
```

but I'm not totally sure how that would look implementation-wise.

goatslacker:
I can prototype out that potential API. Ideally this would ship and we can then support the future stuff on top of this without any (few?) breaking changes.

taion:
Does that API seem good to you, though? Whether or not it's implemented in a backward-compatible manner, following the new API will necessarily involve users re-writing code, since you go from specifying the "fetch" and "get" operations separately to doing so in the same place.

I think it makes a lot of sense to combine those two into the same place for most use cases, since it removes the need to think explicitly about fetching the data when writing the component + containers, but this is also just what I'm used to. It's not unlike how react-resolver works, and I think that's nice.

vincentschroeder:
Is there any progress?

goatslacker:
No, this was dropped.

PR #365, opened 2015-06-28
Adding an empty line before a code section

Satyam:
The Markdown renderer in GitHub seems to accept triple backticks without an empty line above.  The one that generates the site doesn't.


goatslacker:
Thanks!

PR #368, opened 2015-06-29
A debug component

goatslacker:
Creating a debug component for use with React.


Shearerbeard:
So awesome!

goatslacker:
If you want to try this out:

```js
import Debugger from 'alt/utils/Debugger'

<Debugger alt={yourAltInstance} />
```

Shearerbeard:
Trying to test this out but I'm an error that "fixed-data-table" could not be found.

goatslacker:
Ah yes. npm install fixed-data-table 

You'll also want to do that in your app so you can get the CSS. I'm thinking of better ways of packaging that up.

Shearerbeard:
Okay this is seriously cool - I've been playing with it in some test projects and its seems like its running great. Going to wire up a type definition really fast and try it out in a larger app. Thanks!

epeli:
It seems that the css is not active for some reason even thou the style-tag is rendered to the dom. Also the fixed-data-table component does not appear at all but that's probably due to broken css.

The `Start Replay` button seems to work thou.

altdebug

goatslacker:
@epili try `npm install fixed-data-table` I'll work on bundling that in

epeli:
It installed alright. The webpack build crashes if it not. Can't miss it really :)

lenage:
What about `chromeDebug`? could we merge it to Debugger?

goatslacker:
Not sure I understand, chromeDebug enables the chrome debugger whereas Debugger is just a React component you put on your page.

PR #370, opened 2015-07-01
Added util to generate async actions

### LINKED ISSUE #285: Util for generating the 3 action pairs.
Would like a util that takes in a name and generates the 3 actions that you normally use with data sources/async:

Loading, Success, Failure actions.

kristenmills:
Fixes #285

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -81,15 +81,6 @@ class Alt {
     }, actions))
   }
 
-  generateAsyncActions(...actionNames) {
-    const actionsList = actionNames.reduce((list, action) => {
-      list.push(`${action}Loading`, `${action}Success`, `${action}Failure`)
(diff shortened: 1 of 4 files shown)


goatslacker:
Nice, can you add this to utils instead of core?

taion:
Nice. `Starting` might be preferable to `Loading` for e.g. writes to remotes, though.

kristenmills:
Yeah I can make both of those changes.

mattkahl:
I generally like to add a `fooCompleted` as well that's triggered regardless of success or failure. I find it can be especially useful for dependent stores.

kagux:
:+1: for `fooCompleted`/`fooFinished` as we used these a lot.
Is there anything in a way of merging this PR?

PR #373, opened 2015-07-02
Fix DispatcherDebugger commit link in changelog

clessg:
Urgent fix to critical bug.


goatslacker:
Shit. We need to stop the presses!

Thanks for this 😎

PR #379, opened 2015-07-06
Correct document link

tom76kimo:
Hey, noticed the links on Container Components page has broken and fix it.

### DIFF
--- a/guide/container-components/index.html
+++ b/guide/container-components/index.html
@@ -161,8 +161,8 @@
           <div class="sp-horiz-md">
             <h1 id="container-components">Container Components</h1>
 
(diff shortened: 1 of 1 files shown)


jdlehman:
Good catch!  The gh-pages stuff is actually generated. The links need to be updated in [here](https://github.com/goatslacker/alt/blob/master/guides/getting-started/container-components.md).

I'll take a look. It looks like an issue with the generation.

jdlehman:
Ok, I fixed the links and these are working now. Thanks for reporting this :smile:

tom76kimo:
Nice, thank you too!

PR #386, opened 2015-07-07
Let async actions return without dispatching

jokull:


goatslacker:
Nice this just needs a test and it's good to go

test/async-action-test.js
```diff
@@ -0,0 +1,31 @@
...
+import { assert } from 'chai'
+
+const alt = new Alt()
+
+const actions = alt.createActions(class AsyncActions {
+  displayName = 'AsyncActions'
+  fetch() {
+    const obj = {}
+    obj.then = function(){}
+    return obj
```
goatslacker: you can do return Promise.resolve('foo')

goatslacker:
Once it's green...

goatslacker:
Sweet, thanks! Will put out a release.

PR #388, opened 2015-07-08
add connectToImmutableStores.js

loopmode:
even more es7 decorator magic :) allows to easily connect to immutable stores

### DIFF
--- a/src/utils/connectToImmutableStores.js
+++ b/src/utils/connectToImmutableStores.js
@@ -1,0 +1,49 @@
+import connectToStores from './connectToStores';
+/* eslint-disable */
+/**
+ * A Decorator for immutable stores that wraps `alt/utils/connectToStores`.  
+ * Adds the necessary static methods `getStores()` and `getPropsFromStores()`.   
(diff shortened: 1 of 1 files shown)


srph:
Hmm.. isn't this already possible with the `deserialization` hook?

jdlehman:
Yep, this is definitely possible already. @loopmode's util just does some of the boilerplate for you. though depending on if the user has custom deserialization/serialization logic, they might still have to go with the normal `connectToStores` util.

I am not completely opposed to adding this in, but my concern is if we add lots of "specific case" utils then we just bloat alt and make it harder for people to find what they need to do in the docs.  If we make utils a bit more general but configurable, then we can reduce the amount of code to maintain and make it easier to find the tool to do the job.

So after saying all that I would lean towards no on specific utils.

Curious to hear other thoughts though.

goatslacker:
This is a good utility however I don't think I want to be stuck maintaining it. I think its time to move utils/addons outside of alt so these can just exist in user land.

jdlehman:
:+1: to moving them outside of alt.  It looks like that is what redux is doing as well

PR #393, opened 2015-07-10
A proposal for render

goatslacker:


goatslacker:
@taion

taion:
Sorry I've been reading through this in parts but I'm jet lagged/feeling like crap -

First question is that I notice you still separate out the "remote get" logic from the "local get" logic, i.e. `resolveAsync` v `reduceProps` - do you want to eventually add the convenience bit that merges both of those?

foo.js
```diff
@@ -0,0 +1,149 @@
+import Alt from './'
+import React from 'react'
+import Render, { connect } from './utils/Render'
```
taion: You shouldn't be able to import `context` like that - this is only working because of Babel's fallback for when you have a single default export.

src/utils/Render.js
```diff
@@ -115,23 +144,204 @@ function renderWithStrategy(strategy) {
...
+    return Promise.resolve(Element)
+  }
+
+  toDOM(Component, props, documentNode, opts = {}) {
+    return this.getReady(Component, props, opts).then((Element) => {
+      return React.render(Element, documentNode)
+    })
+  }
+
+  static connect(Spec, MaybeComponent) {
```
taion: Per e.g. https://github.com/goatslacker/alt/issues/395 I'm not a fan of making container-generation be a decorator - reason being is that it makes it unclear that you're using a wrapper component, and that e.g. getters on the component class will not be available.

taion: I don't know that there's a direct analogy here, but e.g. I've very intentionally made http://martyjs.org/api/inject-app/index.html usable as a decorator, while http://martyjs.org/api/containers/index.html is not - the former modifies the class in-place and won't break the use cases I referenced above, while the latter adds a container component, and is likely to do so.

foo.js
goatslacker: Ok, fixed this. Will push update soon.

goatslacker:
Regarding resolve vs reduce, yes they need to be separated. Resolve is just about resolving a set of promises before the component is even rendered. Reduce is about what gets sent down to the component.

taion:
I still think it almost always makes more sense to combine the two. I'll prototype something up real quick.

How do I, uhh, run this? I get `TypeError: Cannot read property 'iterations' of undefined` if I do `npm run build` after installing `iso` and `axios` when I hit `localhost:3000` after running `babel-node server.js`.

src/utils/Render.js
```diff
@@ -58,52 +30,109 @@ function usingDispatchBuffer(buffer, Component) {
...
+
+        // clear the buffer and call render again
+        this.promisesBuffer = []
+
+        info.i = i + 1
+        info.time = info.time + (Date.now() - startTime)
+
+        // this takes the rendering out of a Promise context
+        // TODO test this concurrently
+        return setTimeout(() => this.render(alt, Element, info))
```
taion: Something like

```js
        return Promise(resolve => {
          setTimeout(() => resolve(this.render(alt, Element, info)))
        });
```

no?

goatslacker: yeah, fix this line in order to fix the error above!

goatslacker: this isn't necessary actually, im doing this to have a better separation because if a react component throws it'll instead be handled by the Promise which I don't want it to be.

taion: Well the problem is that you resolve the promise with the time out rather than something that unwraps into the expected object.

goatslacker: right, that's silly. What you wrote above should work.

goatslacker:
I don't think they should be combined because they're two different things. Combining them would end up looking a lot like react-resolver/react-transmit but with a flux backing.

taion:
@goatslacker 

That's exactly the approach that Marty takes, though, and I think it's really valuable - it cuts down on a lot of boilerplate and makes the relevant async fetching code much easier to write.

You don't need to explicitly track completed fetches, because those should be handled via the `local` hook. You do need to explicitly track pending and failed fetches, but it's not so bad.

For the average use case where you want to fetch the data once and keep the data around, it's really nice and expressive. I think the only additional thing you might want to do is potentially refresh the data from the client-side in certain instances.

goatslacker:
@cpsubrian had an idea to determine what needs to be passed through by looking at propTypes

taion:
I'm going to do my best to prototype something up as a PR to this branch tomorrow just as a PoC. Didn't get a chance to do so tonight.

cpsubrian:
Note that you could support `React.PropTypes.shape` to even deep-dive, though that might be getting out of control :) Also, if you're using more than just the keys of the propTypes you'd need to have hooks so people using custom types can implement their own logic. For example, I use `react-immutable-proptypes`.

foo.js
```diff
@@ -0,0 +1,182 @@
...
+          </div>
+        ))}
+      </div>
+    )
+  }
+}
+
+@connect({
+  propTypes: {
+    user: React.PropTypes.string.isRequired,
```
goatslacker: container component's propTypes

foo.js
```diff
@@ -0,0 +1,182 @@
...
+  },
+
+  listenTo() {
+    return [UserStore]
+  },
+
+  resolveAsync(props, context) {
+    return context.resolve(UserStore.fetchUsers, props.user, props.repo)
+  },
+
```
goatslacker: no more reduceProps, we can pull the props from the child component's propTypes

foo.js
```diff
@@ -0,0 +1,182 @@
...
+          <div key={user.id}>
+            {this.renderUser(user)}
+          </div>
+        ))}
+      </div>
+    )
+  }
+}
+
+@connect({
```
goatslacker: this is definitely the wrong way to use this, but im just showing an example here.

The right "blessed" way would be to create a new file and do 

```js
import Count from './Count'

const CountContainer = connect(CountDefinition, Count) 

export default CountContainer
```

src/alt/store/StoreMixin.js
```diff
@@ -19,69 +19,38 @@ const StoreMixin = {
...
-    let loadCounter = 0
-
-    const asyncMethods = fn.isFunction(asyncDef)
-      ? asyncDef(this.alt)
-      : asyncDef
-
-    const toExport = Object.keys(asyncMethods).reduce((publicMethods, methodName) => {
-      const desc = asyncMethods[methodName]
-      const spec = fn.isFunction(desc) ? desc(this) : desc
+  fetch(spec) {
```
goatslacker: if we want to merge this in now I should bring back registerAsync and exportAsync and also add fetch. But deprecate the former two.

src/utils/AltIso.js
```diff
@@ -1,26 +1,32 @@
 import Iso from 'iso'
```
goatslacker: we can ignore this file for now as if it doesn't exist.

foo.js
```diff
@@ -0,0 +1,182 @@
...
+    user: React.PropTypes.string.isRequired,
+    repo: React.PropTypes.string.isRequired
+  },
+
+  listenTo() {
+    return [UserStore]
+  },
+
+  resolveAsync(props, context) {
+    return context.resolve(UserStore.fetchUsers, props.user, props.repo)
```
goatslacker: What is this `context.resolve()` magic? Well, if there's an in-flight request then it just uses that promise instead of firing off two requests for the same fetch.

How does it work?

Calls to fetchUsers with the same arguments will be "cached" per context instance and only for a single render lifecycle. If a component's state changes the cache is blown away so this doesn't affect client side re-renders.

foo.js
taion: Huh? What if I want to pass in e.g. `UserStore.getUser(this.props.userId)` as the `user` prop to the child?

goatslacker: You have two options:

1. (explicit) define `reduceProps`

```js
reduceProps(props) {
  return { user: UserStore.getUser(props.userId }
}
```

2. (implicit) set it in the store and make sure it's a propType so it's automatically picked up

```js
this.user = SomeUserObject
```

taion:
ref #400 - a basic prototype of the resolver/transmit-style API, which I think is actually sorta nice. It also doesn't use React context at all, which is a somewhat orthogonal change.

src/utils/Render.js
```diff
@@ -54,80 +56,376 @@ function usingDispatchBuffer(buffer, Component) {
...
+      return this.resolve(
+        new Error('Max number of iterations reached'),
+        html,
+        alt,
+        Element,
+        i
+      )
+    }
+
+    if (info.time > info.timeout) {
```
taion: This is probably not the ideal way to handle timeouts - seems like it'd be better to `Promise.race` for the timeout, or have a flag on the `DispatchBuffer` that gets triggered by the timeout firing; otherwise you end up waiting unnecessarily long.

PR #400, opened 2015-07-14
Context free render

taion:
Obviously not merge-able, but just a trivial PoC, with

1. Marty-style `createContainer` that handles everything
2. App-level interception of fetches to combine fetch resolution and prop reduction
3. Extremely hackish pending fetch tracking (by ID) on the stores
4. Simpler render loop that just does bootstrap -> dispatch -> flush
5. No use of React context anywhere

### DIFF
--- a/foo.js
+++ b/foo.js
@@ -1,6 +1,7 @@
 import Alt from './'
 import React from 'react'
-import Render, { connect } from './utils/Render'
+import createContainer from './utils/createContainer'
+import Render from './utils/Render'
 import axios from 'axios'
 
 const alt = new Alt()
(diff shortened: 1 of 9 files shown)


taion:
This proposed API is definitely less flexible than having the separate entry point for resolving async, _but_ I think in general the only thing you really need the separate `resolveAsync` for is to bypass the local cached value when mounting a component, or when that component's props change, which is a pattern we could support w/e.g. instead of

```js
props: {
  users() {
    return UserStore.getUsers(this.props.user, this.props.repo)
  }
}
```

rather

```js
props: {
  users: {
    value() {
      return UserStore.getUsers(this.props.user, this.props.repo)
    },
    refresh: true
  }
}
```

or something similar, or hoist `refresh` up to a top-level property of the `createContainer` spec.

src/utils/Render.js
```diff
@@ -1,300 +1,58 @@
...
-        html: obj.html,
-        state: obj.state,
-        buffer: obj.buffer,
-        element: obj.element,
-        diagnostics: {
-          iterations: obj.diagnostics.iterations,
-          dispatches: obj.diagnostics.dispatches,
-          time
-        }
+        continuation.onError(result)
```
wonnage: should be onFailure

wonnage:
This might be peculiar to our codebase, but fetches triggered from a deferred Action won't get intercepted by this I think, since they happen outside of the render path - maybe we just shouldn't use them (it's not recommended in the docs)?

goatslacker:
I'm thinking about removing the interception stuff if one of these proposals goes through.

foo.js
```diff
@@ -94,6 +79,15 @@ class Stargazers extends React.Component {
   }
 }
 
+Stargazers = createContainer(Stargazers, {
+  listenTo: UserStore,
+  props: {
+    users() {
```
mull: This'd feel cleaner to me if users() was passed props as a parameter instead of using `this.props`

taion: Hmm, I wonder if you might ever want to access `context` here as well, though.

mull: I see no reason not to pass it along. :)

taion:
Clearly no point leaving this PR open (:

I'm not likely to get a chance to refresh this, but I still think there's something reasonably elegant in here.

PR #401, opened 2015-07-15
Fixed passed value to setState

srph:
is this right? haha


mull:
This fix is indeed correct. :+1:

goatslacker:
It's not setIn like it is now?

mull:
Not sure I get the question @goatslacker but that is definitely how you would work with an Immutable.Map. The previous code replaced the state entirely rather than setting `todos`.

goatslacker:
This is a change set on an old branch. Needs a rebase. Is the current one incorrect?

mull:
Sorry, you're right. This is no longer an issue.

PR #406, opened 2015-07-16
This article tells a very good story of why Alt  over other flux frameworks

amunda:
Added under examples but we can even create a separate heading eg. usecases or companies using Alt or anything else


goatslacker:
Thanks :D

Issue #366, opened 2015-06-29
Snapshot / Bootstrap with Immutable.js AltStores?


mwise:
First off, thanks for Alt - I'm just starting to integrate it and it's been a breeze.

I'm having some difficulty getting  `alt.takeSnapshot()` and `alt.bootstrap()` to work when using `AltStore` instances and the`alt/utils/ImmutableUtil` wrapper function. I *think* I have followed the directions for its use as per https://github.com/goatslacker/alt/blob/master/docs/utils/immutable.md. When I call `alt.takeSnapshot()`, the returned string looks OK. That is, it contains the data in the format prescribed at http://alt.js.org/docs/bootstrap/. However, when I pass that string back into `alt.bootstrap()`, crazy things happen to the data and my application state goes bye-bye.

I put together a branch of the alt-tutorial repo that illustrates the weirdness that I'm seeing:  https://github.com/goatslacker/alt-tutorial/compare/master...mwise:immutable-stores

To see the issue, run `npm start`, then click some favorites, then click 'Take snapshot', then click 'Restore snapshot' and watch all the data go away.

The "crazy things" that happen to my data are a little tough to explain. I did some debugging (in my actual app, not in the above linked repo), and it looked like the call to `Immutable.fromJS()` at https://github.com/goatslacker/alt/blob/master/src/utils/ImmutableUtil.js#L19 was being passed the correct objects. However - and this is kind of the crazy part - the calls to `Immutable.fromJS()` seemed to be returning results as if they were reducing the values being returned. That is, if my snapshot string was: 

```
'{"AccessTokenStore":{"accesstokenvalue":"ACCESSTOKEN"},"ConfigStore":{"configvalue":"CONFIG"}}'
```

I would see successive calls to `Immutable.fromJS()` that looked like this:

 ```
// call 1:
Immutable.fromJS({ accesstokenvalue: "ACCESSTOKEN" })
// returns an Immutable.Map such that:
AccessTokenStore.getState().toJS() === { "accesstokenvalue": "ACCESSTOKEN" }

// call 2:
Immutable.fromJS({ configvalue: "CONFIG" })

// returns an Immutable.Map such that:
ConfigStore.getState().toJS() === { 
  "accesstokenvalue": "ACCESSTOKEN", 
  "configvalue": "CONFIG" 
}

// WAT?!
```

[...]

PR #411, opened 2015-07-22
Fix bootstrap with immutable store

ahanriat:
Related to #366 

Hi guys, a small PR to fix an issue when using Immutablejs and `alt.bootstrap`. I've just added a dummy `onBootstrap` method in `immutableUtils` and used it in the `setAppState` function to avoid side effect on Immutablejs.

Please let me know if you need anything else.

Thanks a lot for your work !!


goatslacker:
This needs tests. Also why can't we use onDeserialize for this case? What happens currently?

There are tests for bootstrapping with immutable.js

jdlehman:
I might need to dig in more, but I am using Immutable in production with no issues bootstrapping my data.  `bootstrap` expects stringified JSON and after parsing this JSON, I believe that the `onDeserialize` is called (which is already provided in the Immutable utils) and takes care of making this POJO an immutable data structure and sets it to the state of the store.

However, I might not be understanding the actual issue.

ahanriat:
Thanks a lot for your quick answer !
The issue appears when you want to bootstrap a store which have an empty map for state (which is not the case with existing tests) ie:
```
this.state = Immutable.Map()
```

Then if you boostrap it, you won't be able to pass the test:
```
      assert.deepEqual(Immutable.fromJS({}).toJS(), {}, 'the prototype of Immutable.Map is ok')
```

I'll review my PR to see if I can use the `onDeserialize` and add some tests :wink:

jareware:
I've also ran into this while using `Immutable.Record` as my `this.state` (which is actually implemented under the hood as `Immutable.Map`).

ahanriat:
@goatslacker I've added some tests to check the behaviour of ImmutableJS when bootstrapping an empty store ;)

Hope this will help

jacobangel:
Just chiming in, ran into this issue when State was an Immutable.Set. 

Seems the delete portion of `setAppState` blows out a few shared internals, which in turn breaks all new empty sets.

jareware:
Yes, it also breaks with Lists, Records and basically anything other than a Map... and frankly I'm not quite sure why even the Map works, it doesn't look like it should.

I added some tests documenting these failures in #452.

jareware:
Would it be enough to just use the return value of `onDeserialize()` as the new state, without any further mangling?

mull:
^ using the return value of `onDeserialize` is what I do.

jareware:
@ahanriat how do you do that, though? Isn't the root of the problem that regardless of what `onDeserialize()` returns it's subjected to [this mangling](https://github.com/ahanriat/alt/blob/master/src/alt/utils/StateFunctions.js#L11)?

jareware:
For the record, this is what I've patched in, and works fine:

```js
if (config.onDeserialize) {
  config.setState.call(store, state, config.onDeserialize(value) || value);
} else {
  fn.eachObject(function(k) {
    return delete state[k];
  }, [state]);
  fn.assign(state, obj[key]);
}
```

mull:
@jareware sorry it looked like I had misunderstood the initial issue. Oddly enough I do use snapshotting & bootstrap with alt 0.17.1 but I don't have this problem. However, I only use Maps & Sets.

goatslacker:
Merged your test in: https://github.com/goatslacker/alt/commit/1a812652948c5cc06e4d36fccb32acf9f9b6e59d

Thanks!

beeb:
I hate to reopen this but I'm having a problem with boostrap and using a Record as my state. The state that is created with `alt.bootstrap` is a `Map` whereas my original data was a `Record`. I'm using the immutable utils on the store and everything else works fine.

```
const State = Record({
  localStorage: false,
  lastSave: '',
  loading: false,
  error: false,
  errorMessage: '',
  filename: '',
  arrName: '',
  totalLines: 0,
  byteArray: new Uint8Array(),
  header: new Uint8Array(),
  items: List(),
  itemsById: Map() //custom Record instances
})
```

For now I'm using https://github.com/glenjamin/transit-immutable-js and do the serialization myself.

PR #415, opened 2015-07-26
Make the documentation more succinct

gajus:
* Include doc refs
* Use ES6 consistently (let should be used instead of var)

### DIFF
--- a/guides/getting-started/actions.md
+++ b/guides/getting-started/actions.md
@@ -7,26 +7,30 @@ permalink: /guide/actions/
 
 # Creating Actions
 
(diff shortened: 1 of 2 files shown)


guides/getting-started/actions.md
```diff
@@ -7,21 +7,23 @@ permalink: /guide/actions/
...
+
+Export the class using [`createActions`](http://alt.js.org/docs/createActions/#createactions).
 
 ---
 
 `actions/LocationActions.js`
 
 ```js
-var alt = require('../alt');
+let alt = require('../alt');
```
mull: const?

PR #421, opened 2015-07-28
Added a new link to an article about Altjs and React

richard-lopes:


jdlehman:
Great find! I wonder if we should have this in a separate section called `Articles` or `Tutorials` or something since it provides more context/guidance than a typical github repo.

Maybe it is fine here for now and we can move it later if we get more articles (if that ever happens).

rlopes:
+1

goatslacker:
Thanks, great article.

richard-lopes:
Thanks ! We plan to write new content around Alt.js including testing.

PR #427, opened 2015-07-30
Support defaultProps in connectToStores

pstoica:
Noticed you could never get defaultProps in getPropsFromStores. This should fix it!


jdlehman:
Can you add tests for this?

pstoica:
Will do.

mull:
@pstoica still planning on doing that?

PR #429, opened 2015-07-31
Update createAction docs

jareware:
Clarified the return-to-dispatch semantics, and its exceptions. Added examples.


goatslacker:
:+1:

PR #433, opened 2015-08-02
Enable hot loading for stores if not in production

### LINKED ISSUE #408: makeHot doesn't accept possible Store parameters
I was trying to port code such as `this.store = alt.createStore(NoteStore, storeName, this.actions);` to use `makeHot`. Sadly it looks like `makeHot` doesn't allow parameters to be passed to Store. I expect that for this to work you would need to do something like:

```javascript
function makeHot(alt, Store, name = Store.displayName, parameters) {
  if (module.hot) {
    module.hot.dispose(() => {
      delete alt.stores[name]
    })
  }

  return alt.createStore(Store, name, parameters)
}

export default makeHot
```

Maybe this was just a simple oversight? I can provide a PR if you want but given it's so trivial change, perhaps just add it there yourself it this makes sense to you.

bebraw:
I tested this against an app of mine and it worked like a charm. Of course you may want to try it too before merging.

You might want to deprecate `makeHot` at some point as this little change makes it redundant.

Closes #408.

### DIFF
--- a/src/alt/index.js
+++ b/src/alt/index.js
@@ -46,6 +46,15 @@ class Alt {
   }
 
   createStore(StoreModel, iden, ...args) {
+    // allow hot loading for development usage
+    if (process.env.NODE_ENV !== 'production') {
+      if (module.hot) {
+        module.hot.dispose(() => {
+          delete this.stores[name]
+        })
+      }
+    }
+
     let key = iden || StoreModel.displayName || StoreModel.name || ''
     store.createStoreConfig(this.config, StoreModel)
     const Store = store.transformStore(this.storeTransforms, StoreModel)


src/alt/index.js
```diff
@@ -46,6 +46,15 @@ class Alt {
   }
 
   createStore(StoreModel, iden, ...args) {
+    // allow hot loading for development usage
+    if (process.env.NODE_ENV !== 'production') {
+      if (module.hot) {
```
goatslacker: I was playing with this today, would this work better?

```js
if (module.hot) delete this.stores[name]
```

bebraw: Yeah, that's possible. I just ported the code from the current hot loading utility and didn't really think about it in detail.

Feel free to do whatever changes you feel necessary. One problem with the current PR is that it introduces a code path that's not tested (-> lower coverage).

goatslacker: Right on, I'm going to merge this today.

goatslacker:
Blocked trying to add tests to this.

goatslacker:
8485eee7e592421eefb1596f469eafd43e9cb051

PR #437, opened 2015-08-03
Use custom onSerialize method if present when saving initial snapshot

chetstone:
I was having trouble trying to create a store that has a reference to Firebase. I can do it with `createUnsavedStore()`, but I'm trying to use `altmanager`, and that dosen't work with unsaved stores.

The error when calling `createStore()` was `Uncaught TypeError: Converting circular structure to JSON`.

I created a custom serializer but it seems that the initial snapshot when `createStore()` is called does not use the custom serializer if present. This fixes that. 

The fix caused one test to fail, because the assumption was that a newly created store has never been serialized. This was actually incorrect, because it's always serialized at creation, it just wasn't detectable before. So I just removed the test.

### DIFF
--- a/src/alt/utils/StateFunctions.js
+++ b/src/alt/utils/StateFunctions.js
@@ -30,8 +30,12 @@ export function snapshot(instance, storeNames = []) {
 }
 
 export function saveInitialSnapshot(instance, key) {
+  const store = instance.stores[key]
+  const { config } = store.StoreModel
+  const customState = config.onSerialize &&
+    config.onSerialize(store.state)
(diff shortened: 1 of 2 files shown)


goatslacker:
you probably shouldn't keep references to firebase as store state.

Here's something you can try:

```js
class FirebaseStore {
  constructor() {
    this.firebaseinst = new Firebase();

    this.state = { myState: true, helloWorld: 20, bananas: 'donkeys' }
  }
}
```

chetstone:
That works. Sweet!

Thanks

PR #439, opened 2015-08-04
Fix ImmutableUtils docs

AsaAyers:
After I fixed the `properties` that should be `prototype` I found that `this.setState(this.state.todos.set(id, todo));` was wrong. That ends up replacing `this.state` with the modified version of `this.state.todos`. I have no prior experience with Alt, this is just the solution I backed into while trying to figure out and fix the `this.setState` call. Since "Alt has first-class support for immutable data structures via the ImmutableUtil" it does feel like I'm missing a better solution.


goatslacker:
:fire: Thanks!

PR #448, opened 2015-08-07
Set state in mixinContainer after mounting

alexlande:
Hey! Thanks for Alt, it's rad. This feels sort of dumb but wanted to run it by you. 

I hit this edge case today-- I have a component that renders a Google map. After mounting it persists a reference to the map instance in a store. My components are inside an `AltContainer`, with a structure like this:

```
Page
  > AltContainer
    > Map
    > OtherStuff
```

The AltContainer mounts after the Map does, which means that in between the container's `getInitialState` and subscribing to my store, the store's state changes, and the container isn't aware of it.

This PR adds an `altSetState` call after the container mounts, which solves this problem by causing an additional render, which is a bummer.

My current workaround in my app is to defer the action that the map fires, which makes me feel bad about myself but works.

### DIFF
--- a/components/mixinContainer.js
+++ b/components/mixinContainer.js
@@ -46,6 +46,7 @@ function mixinContainer(React) {
 
     componentDidMount: function () {
       this.registerStores(this.props)
+      this.altSetState(this.props)
     },
 
     componentWillUnmount: function () {
--- a/test/store-listener-component-test.js
+++ b/test/store-listener-component-test.js
@@ -327,7 +327,7 @@ export default {
         </AltContainer>
       )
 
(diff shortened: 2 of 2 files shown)


goatslacker:
We shouldn't call this method, either do a this.forceUpdate or better yet add a hook so we can call something when the component mounts? Similar to componentDidConnect on connectToStores

goatslacker:
Added an onMount handler 50eb8041bf5fb9e2bbbd1ae8f28fd84f6083740e

goatslacker:
You use it like:

```js
<AltContainer onMount={callback} />
```

alexlande:
That's awesome, thanks :metal:

PR #449, opened 2015-08-08
ImmutableUtil Store's setState should merge state

droberts84:
Currently when using `ImmutableUtil` to use Immutable data in a Store, calling `setState` replaces the current state with the argument passed, which is unexpected as it does not behave like `setState` in React components.

This PR changes `setState` to merge the new provided state with the current state.

There are a few edge cases that become odder with this change, such as removing a key from an `Immutable.Map` (see tests), but overall I think it is a win and much less confusing to users.

Note that this could potentially be a breaking change to current users of `ImmutableUtil`.

### DIFF
--- a/docs/utils/immutable.md
+++ b/docs/utils/immutable.md
@@ -57,8 +57,9 @@ alt.createStore(immutable(TodoStore));
 A few things to note about immutable stores about this approach:
 
 * You use `this.state` to create your state rather than assigning directly to instance properties.
-* You specify your own Immutable data structure you wish to use, as long as it supports `merge`. In this example we're using Map.
-
+* It is highly recommended that you use an `Immutable.Map` as the top level state.
+  * `merge` will be called whenever calling `this.setState`, so other structures can be used as long they support `merge`
+  * WARNING: `Immutable.List` will likely not behave as you expect when calling `setState` in the current version of `Immutable`.  See [this issue](https://github.com/facebook/immutable-js/issues/406) for more information.  Instead it is recommended to nest the `List` in a `Map` as shown above.
 
 Changing the state of your ImmutableStore is a bit different from using a regular store:
 
--- a/docs/utils/immutable.md
+++ b/docs/utils/immutable.md
@@ -30,9 +30,10 @@ class TodoStore {
   static displayName = 'TodoStore'
 
   constructor() {
-    this.state = {
-      todos: Immutable.Map({})
-    };
+    this.state = Immutable.Map({
+      todos: Immutable.List([]),
+      loading: false
+    });
   }
 }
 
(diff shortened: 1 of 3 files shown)


mull:
I wrote my own decorator to do exactly this. I agree with the behaviour. It's a breaking change though of course since it forces you to use Immutable.Map for your state instead of a normal JS object.

droberts84:
I considered doing a check to see if the `merge` function exists before calling it, which would allow you to use a normal JS object in your state.. but I wasn't sure if it was worth the potential performance hit, plus it could still be a breaking change (except without an error) if they were expecting the state to be fully replaced, as it did in previous versions.

cayuu:
I too currently run a monkey patched decorator to get this behaviour. However this current PR will pass through .merge mashing of lists: https://github.com/facebook/immutable-js/issues/406 - the decorator I use applies the `.mergeWith` "fix" in the first comment on that issue. Given that Immutable looks like they're adopting the fix in the next major version bump, I'd suggest to maybe ponyfill the fix or document the hell out of the current wonky List overwriting.

droberts84:
Interesting.  I don't typically use Lists so I wasn't aware of that
behavior, but that definitely is not desired.  That said, I'm hesitant to
add code in Alt to work around an issue with a specific Immutable structure
(especially since they plan to fix it in the next major release).

What do others think?

docs/utils/immutable.md
```diff
@@ -30,9 +30,10 @@ class TodoStore {
   static displayName = 'TodoStore'
 
   constructor() {
-    this.state = {
-      todos: Immutable.Map({})
-    };
+    this.state = Immutable.Map({
+      todos: Immutable.List([]),
+      loading: false
```
goatslacker: whats this loading all about?

goatslacker: just as an example?

src/utils/ImmutableUtil.js
```diff
@@ -3,7 +3,7 @@ import Immutable from 'immutable'
 function immutable(StoreModel) {
   StoreModel.config = {
     setState(currentState, nextState) {
-      this.state = nextState
+      this.state = currentState.merge(nextState)
```
goatslacker: this works well but then limits us to only being able to use Map

cc @jareware

jareware:
This would work, and it has the nice property of maintaining whatever type state already was.

Still, e.g. #411 actually prevents you from using anything other than a Map, so at least until that's fixed it's a bit of a moot point. :)

docs/utils/immutable.md
droberts84: Yes `loading` was just meant to be an example of a random variable you might include in state.  Any suggestions to clarify this?

goatslacker:
I'd like to be able to use something other than Map. #485 handles all cases

PR #452, opened 2015-08-11
Bootstrapping ImmutableUtil with Records

jareware:
Bootstrapping an immutable Store that uses `Immutable.Record` as its state doesn't seem to work. The "bootstrapping record" case documents this as a failure.

My guess is there's something misbehaving in `setAppState()` in StateFunctions.

### DIFF
--- a/test/immutable-stores.js
+++ b/test/immutable-stores.js
@@ -132,6 +132,57 @@ export default {
 
       assert(store.getState().get(0) === 1)
     },
+    
+    'using record'() {
+      const MyRecType = Immutable.Record({ x: 123 });
+      const alt = new Alt()
+      const action = alt.generateActions('addX')
+      const store = alt.createStore(immutable({
+        bindListeners: { addX: action.addX },
+        state: new MyRecType(),
+        addX() {
+          this.setState(this.state.set('x', 456))
+        }
+      }), 'RecordImmutableStore')
+
+      assert(store.getState().x === 123, 'store contains expected initial value 123')
+
+      action.addX()
+      assert(store.getState().x === 456, 'store has updated after action')
+    },
+
+    'bootstrapping map'() {
+      const alt = new Alt()
+      const action = alt.generateActions('addX')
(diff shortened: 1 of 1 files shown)


jareware:
Right, it looks like the issue is that ImmutableUtil uses `fromJS()` to revive the state, which expectedly yields an `Immutable.Map`. But instead of using e.g. the Store's `config.setState()`, which could for instance merge them, the state is patched directly by `setAppState()`. I'm actually not even sure why this works for Maps... :)

goatslacker:
Sick. Ok I'll take a look at this.

jareware:
Awesome, thanks!

goatslacker:
Ok so its because of this line: https://github.com/goatslacker/alt/blob/master/src/utils/ImmutableUtil.js#L19

which automatically turns things into a Map

jareware:
Yes!

> Right, it looks like the issue is that ImmutableUtil uses `fromJS()` to revive the state, which expectedly yields an `Immutable.Map`. But instead of using e.g. the Store's `config.setState()`, which could for instance merge them, the state is patched directly by `setAppState()`.

The thing is, though, that you can't customize this even if you wanted to, because `setAppState` will break pretty much anything except a POJSO.

Please see #411 for the root cause, and for others who are struggling with this! :)

goatslacker:
I'm aware of 411 but I'm not sure if I want to add more API...

Maybe what we need is just a way to control how state is set.

jareware:
I think that's exactly what we need. The [solution I outlined](https://github.com/goatslacker/alt/pull/411#issuecomment-134935338) does this, by giving complete control to the store, via its config methods.

I agree about not adding more API, and you'll note that I have used the existing ones for that. The resulting change in semantics is minimal, and would give a lot more than it would take away.

What do you think @goatslacker?

goatslacker:
Let me load that one in a separate branch.

goatslacker:
#485 handles this

jareware:
Wow, excellent news, thanks!

Issue #438, opened 2015-08-03
Confused sample code on view listen to  store


variable:
On this page:

http://alt.js.org/guide/view/

> But then we’ll want to listen to changes once the state in the store is updated. In your react component > on componentDidMount you can add an event handler using LocationStore.listen.

    LocationStore.listen(this.onChange);

This seems reversed while reading it naturally, the code makes me read that "the store listen to the view change."

variable:
It makes more sense with something like

    LocationStore.addListener(this.onChange);

goatslacker:
It definitely does. This API was [borrowed from RefluxJS](https://github.com/reflux/refluxjs#listening-to-changes-in-data-store)

I've thought about changing it to `subscribe`/`destroy` but I think that :ship: has sailed at this point.

variable:
I think at least putting a note on the documentation and the tutorial will help a lot.

goatslacker:
It already exists right towards the bottom of this section: https://github.com/goatslacker/alt#stores

and also here: http://alt.js.org/docs/stores/#storelisten

PR #458, opened 2015-08-20
added addListener and removeListener

variable:
This is for the issue https://github.com/goatslacker/alt/issues/438

I have added 2 alias functions to help better code readability while retaining compatibility

    LocationStore.addListener(this.onStoreChange);

### DIFF
--- a/src/alt/store/AltStore.js
+++ b/src/alt/store/AltStore.js
@@ -86,6 +86,14 @@ class AltStore {
     this.transmitter.unsubscribe(cb)
   }
 
+  addListener(cb){
+    return this.listen(cb)
+  }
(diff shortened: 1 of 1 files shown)


goatslacker:
If we're keen on adding new methods then we should remove the old ones.

And if we're doing that then I think it should be named 'subscribe' and 'destroySubscription' or something similar to that.

I don't want to add alias methods because this grows the library in size and complexity.

variable:
But 'subscribe' is as confusing as 'listen', LocationStore doesn't not subscribe to view changes. Adding alias will allow backward comparability so other people don't need to change the method names.

PR #460, opened 2015-08-21
Copy child context types when creating storeconnection

wonnage:
This makes using instances of alt by passing them through React context easier. StoreConnection doesn't have any contextTypes defined, so when it calls getStores/getPropsFromStore and passes in props/context, context will always be empty. To fix this, we can copy the contextTypes your component would've received, if it weren't wrapped, onto StoreConnection itself.

### DIFF
--- a/src/utils/connectToStores.js
+++ b/src/utils/connectToStores.js
@@ -89,6 +89,9 @@ function connectToStores(Spec, Component = Spec) {
       )
     }
   })
+  if (Component.contextTypes) {
+    StoreConnection.contextTypes = Component.contextTypes
+  }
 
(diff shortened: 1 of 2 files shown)


goatslacker:
If you use the object passing form of connectToStores this works, maybe we should get rid of the static method thing.

```js
connectToStores({
  getPropsFromStores() { }
})
```

wonnage:
Not totally sure what you mean - if I have `connectToStores(TestComponent)`, I want the context passed to `TestComponent.getPropsFromStores()` to respect TestComponent's `contextTypes`. I don't think that can happen without this change.

PR #464, opened 2015-08-24
Return local's value rather than undefined

mull:
This enables data sources to be used as promise chains without having to do cumbersome conditionals to check whether the data source method will return a Promise or undefined. Currently any value returned from `local` will never be return to the caller.

That fact is a big issue as promise based workflows, such as having to load certain things in order, become cumbersome to write. Consider this:

```
let firstNotification, author;
NotificationStore.fetch().then(() => {
  firstNotification = NotificationStore.first();
  return UserStore.find(firstNotification.user_id);
}).then(() => {
  // Yay we're done;
});
```

If the data source's `shouldFetch` returns false, then this chain breaks immediately. Given that the lack of a return statement and unspecified behaviour I wouldn't consider this a breaking change.

I would still like to add tests but before wrapping my head around how to write that I'd rather see if this'll get accepted or not.

### DIFF
--- a/src/alt/store/StoreMixin.js
+++ b/src/alt/store/StoreMixin.js
@@ -72,6 +72,7 @@ const StoreMixin = {
         } else {
           // otherwise emit the change now
           this.emitChange()
+          return value
         }
       }
 


goatslacker:
I'm going to remove the emit change from this eventually but this is a good change for now, thanks!

mull:
Thanks for the merge @goatslacker, any chance you could bump the version number & release as well? For some reason I seem unable to use github master to install alt.

goatslacker:
I'll make a release pretty soon, there are other changes that I'd like to get out as well.

PR #465, opened 2015-08-26
Enable installing via git

mull:
To make testing new things easier, before a new version is released, it would be tremendously helpful to be able to install alt from the git repository.

Unfortunately `lib` and `utils` are both in .gitignore, making this impossible. The only around that is to fork locally and specify `file:/path/to/alt` in your package.json.

Now this obviously creates a bit of noise. When merging things one will have to more or less ignore anything in lib & utils and just re-build once the other files are resolved. I see no other way around this issue though.

Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,6 @@ coverage
 node_modules
 npm-debug.log
 test/browser/tests.js
-lib
-utils/*
 flux.js
 flux-build.js
 flux-build.min.js
--- a/lib/actions/index.js
+++ b/lib/actions/index.js
@@ -1,0 +1,92 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {
+  value: true
+});
+
+var _createClass = (function () { function defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ('value' in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } } return function (Constructor, protoProps, staticProps) { if (protoProps) defineProperties(Constructor.prototype, protoProps); if (staticProps) defineProperties(Constructor, staticProps); return Constructor; }; })();
+
+exports['default'] = makeAction;
+
+function _interopRequireWildcard(obj) { if (obj && obj.__esModule) { return obj; } else { var newObj = {}; if (obj != null) { for (var key in obj) { if (Object.prototype.hasOwnProperty.call(obj, key)) newObj[key] = obj[key]; } } newObj['default'] = obj; return newObj; } }
+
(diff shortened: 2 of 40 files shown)


goatslacker:
Ouch quite a bit of noise, perhaps we could just publish things to npm prematurely?

Or create a separate package that contains the compiled output?

taion:
This is also just the wrong way to do it. See https://github.com/npm/npm/issues/3055.

If we needed to hack around this urgently, I'd just add a custom `postinstall` script that runs a build only if the built artifacts aren't available.

mull:
@taion that wouldn't work as installing from git doesn't install devDependencies... pretty sure it doesn't, maybe a newer version of npm than I am on does. If newer npm does then that would be awesome! I don't like the way this PR looks either to be honest.

taion:
Why do you need `devDependencies` if you're installing this as a dependency of something else?

mull:
To run the build script you would need the devDependencies right? Since installing via got does not include them then the postinstall would fail.

taion:
Oh, hah, you're right. I suppose you could make the post-install script do both, but... eh.

goatslacker:
Yeah I thought about putting the dev dependencies into dependencies but I don't like that idea either because then babel gets installed with every `npm install alt`

goatslacker:
I really like just publishing to npm the early releases or just creating a separate repo.

taion:
Well, on second thought, having the post-install script install the dev dependencies if it notices that it got installed without the built artifacts doesn't sound that bad. It would address some legitimate use cases with developing against GH branches.

mull:
```
#!/bin/sh
# Build the project unless build artifacts are already present.

if [ ! -d "./dist" ]; then
  npm run build
fi
```

```
{
  "postinstall": "sh postinstall.sh"
}
```
Something along these lines? I tried it out actually & it works perfectly. Does not work on WIndows though. Could perhaps be written as a node script rather than a shell script.

taion:
I think it'd have to be

```shell
npm install
npm run build
```

To get around the dev dependency thing you mentioned earlier. But otherwise something like this!

mull:
Do you think we need to bother with platforms that don't support bash?

goatslacker:
Yes we should bother. There are people that use windows and such

mull:
Indeed. Can't say I have much of a clue for what to do on anything else than osx/linux. I'm thinking whether it's perhaps easier to just set up CircleCI to push to a separate repository that includes the build files, like was suggested earlier.

mull:
I'm working on the alternative solution to this. Will provide a PR once it's ready. Wouldn't be a bad idea to create an `alt` organisation where we could host the repository & also create org tokens.

goatslacker:
https://github.com/altjs

taion:
BTW, React Router puts this in the post-install script. Took a few iterations to get entirely right, but it seems to largely work in a lot of cases now: https://github.com/rackt/react-router/blob/master/npm-scripts/postinstall.js

mull:
Ugh sorry for the absolute inactivity from me :( react-router's solution looks pretty nice :)

PR #468, opened 2015-08-27
Only delete/update if its a POJO

goatslacker:
cc @jareware I think this might fix the immutable issue


goatslacker:
At least part of it

goatslacker:
This works but I'm not proud of it.

jareware:
Looks like exactly what's needed! :+1:

PR #473, opened 2015-09-06
Update eslint & CONTRIBUTING.md. Also fix a few lint errors.

mull:
Had a hard time getting up & running on circle to set up the github-clone with lib & util files for git installing. Tracked that down to `prepublish` being run before `npm install` when installing locally. This is really annoying as it depends on global installs, but I guess it's something we have to live with.

I went ahead and updated the eslint dependencies and also added which files need to be installed globally to CONTRIBUTING.md to be able to install locally.


mull:
Would it be alright to add a few never iojs versions to travis.yml? With 4.0.0-rc being out 1.8.1 seems a tad old. :)

mull:
Interesting lint errors referring to lines that don't really reflect the errors. Perhaps we could also move to the new Travis infrastructure while at it? Wouldn't mind setting this up.

goatslacker:
Closing in favor of #486

PR #474, opened 2015-09-06
Make the IFs less verbose on BindActions

af7:
Not important change, just less lines (and maybe slightly more performant?)

### DIFF
--- a/src/alt/store/StoreMixin.js
+++ b/src/alt/store/StoreMixin.js
@@ -144,14 +144,9 @@ const StoreMixin = {
           `You have multiple action handlers bound to an action: ` +
           `${action} and ${assumedEventHandler}`
         )
-      } else if (this[action]) {
-        // action
-        handler = this[action]
-      } else if (this[assumedEventHandler]) {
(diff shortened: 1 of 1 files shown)


taion:
That function still reads a bit funny. Wouldn't it be better off as e.g.:

```js
      const matchFirstCharacter = /./
      const assumedEventHandler = action.replace(matchFirstCharacter, (x) => {
        return `on${x[0].toUpperCase()}`
      })

      if (this[action] && this[assumedEventHandler]) {
        // If you have both action and onAction
        throw new ReferenceError(
          `You have multiple action handlers bound to an action: ` +
          `${action} and ${assumedEventHandler}`
        )
      }
      
      const handler = this[action] || this[assumedEventHandler]
      if (handler) {
        this.bindAction(symbol, handler)
      }
```

goatslacker:
:+1: to @taion's change

af7:
@taion Indeed, even better, I will update the PR with your suggestion later on the day

af7:
Ok, just updated.

PR #476, opened 2015-09-08
Copy contextTypes from component to StoreConnection in connectToStores

alexesDev:
Without this fix code like this doesn't work:
```js
@connectToStores
export default class Header extends React.Component {
  static contextTypes = {
    flux: React.PropTypes.any.isRequired
  }

  static getStores(props, context) {
    return [context.flux.getStore('application'), context.flux.getStore('cart')];
  }
}
```
...or is it better to separate the context of Component and StoreConnection?


mull:
LGTM. Useful for wrappers that might need to react to a load event / prop value and redirect to another page (looking at `react-router`.) Get's my :+1:

goatslacker:
Can you add a test for this?

goatslacker:
nvm merged #460

PR #485, opened 2015-09-11
Fix setState so it handles values

goatslacker:
When using a store that has state that is a value and not an object we still want to be able to use this.setState

### DIFF
--- a/src/alt/store/index.js
+++ b/src/alt/store/index.js
@@ -26,12 +26,6 @@ function doSetState(store, storeInstance, state) {
 }
 
 function createPrototype(proto, alt, key, extras) {
-  proto.boundListeners = []
-  proto.lifecycleEvents = {}
-  proto.actionListeners = {}
-  proto.publicMethods = {}
-  proto.handlesOwnErrors = false
-
   return fn.assign(proto, StoreMixin, {
     displayName: key,
     alt: alt,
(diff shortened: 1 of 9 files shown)


src/alt/store/index.js
```diff
@@ -53,7 +53,13 @@ export function createStoreConfig(globalConfig, StoreModel) {
         return state
       }
     },
-    setState: fn.assign
+    setState(currentState, nextState) {
+      if (Object.prototype.toString.call(nextState) === '[object Object]') {
```
taion: Maybe pull this out into a `fn.isPlainObject`? You use it e.g. at https://github.com/goatslacker/alt/blob/44d581401bcf03d531537a70285edf28fcebfc4b/src/alt/store/index.js#L50 as well.

goatslacker: yeah fn.isPlainObject sounds like a good idea. Not sure if I should merge #468 or not. It's a bit dirty to add a property to alt stores and check for them.

taion: With something like #344 it could be a non-issue as you could potentially have a separate overriding implementation of `getState` for stores holding immutable data.

goatslacker: Yeah, but #344 opens a whole can of worms in how state is managed and exposes a lot of internals. I'm still conflicted on it.

goatslacker:
Fixes #468

PR #488, opened 2015-09-14
when update the current thread ,the last message in the thread can not update.

songquanwang:
when update the current thread ,the last message in the thread can not update.


goatslacker:
Is this also in Facebook's chat example?

songquanwang:
Yes,this is in the  Facebook's chat example.

goatslacker:
Can you point me to where? I can't find this in Facebook's chat example.

PR #498, opened 2015-09-29
Allow dispatching actions with action creators

goatslacker:
This allows the use of action creators which is good for using instances since now you can import actions directly and then just dispatch them.

```js
const increment = {
  id: 'my/increment/action',
  dispatch(x) {
    return x
  }
}

alt.dispatch(increment, 2) // will call my/increment/action with a payload of 2
```

### DIFF
--- a/src/alt/actions/index.js
+++ b/src/alt/actions/index.js
@@ -34,7 +34,7 @@ export default function makeAction(alt, namespace, name, implementation, obj) {
     // async functions that return promises should not be dispatched
     if (!newAction.dispatched && result !== undefined && !fn.isPromise(result)) {
       if (fn.isFunction(result)) {
-        result(dispatch)
+        result(dispatch, alt)
       } else {
         dispatch(result)
       }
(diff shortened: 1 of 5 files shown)


src/alt/actions/index.js
```diff
@@ -34,11 +34,21 @@ export default function makeAction(alt, namespace, name, implementation, obj) {
...
       } else {
         dispatch(result)
       }
     }
+
+    if (!newAction.dispatched && result === undefined) {
+      /* istanbul ignore else */
+      /*eslint-disable*/
+      if (typeof console !== 'undefined') {
+        console.warn('An action was called but nothing was dispatched')
```
taion: Consider maybe https://www.npmjs.com/package/warning?

goatslacker: Ooo, good find

taion: I've also gotten good mileage out of invariant, but that's not applicable here.

Issue #501, opened 2015-10-01
ActionsClass usage


wonnage:
I have some information in my stores, that my actions need to pass when creating an API client object. I'm using alt instances, so the API constructor has no access to these stores - I have to pass the data in Intuitively, I'd do something like this:

````
class MyAction {
  constructor() {
    this.apiClient = new ApiClient(this.alt.stores.AuthStore.getStuff());
  }
  
  makeRequest(params) {
    this.dispatch(params)
    this.apiClient.get('/endpoint', params)
  }
}
````

Unfortunately, this doesn't work, because the way Alt creates actions rips all the methods off of this class and binds them to an AltAction instead. Do you have any suggestions on alternatives, or maybe how we could improve how makeActions works with a class? As it stands, the classes don't behave like classes at all, and might as well be object hashes.

goatslacker:
> As it stands, the classes don't behave like classes at all, and might as well be object hashes.

+1

let me think about this.

goatslacker:
ok there are two options here.

remove this.dispatch from alt or remove action classes altogether.

taion:
Hey! I made the exact same complaint here: https://github.com/goatslacker/alt/issues/359 :D

I had a few proposals there toward the end of the thread, all of which would have been major breaking API changes, but probably getting rid of action classes would be the simplest approach by far.

goatslacker:
#503

goatslacker:
Will be releasing this at some point soon.

Issue #359, opened 2015-06-27
underdash: Actions


taion:
ref #337 

From https://github.com/goatslacker/alt/issues/337#issuecomment-115830538:

---

Actions/constants in Alt generally look very nice to work with due to the reduced boilerplate, but my main concern is that `Actions` classes implement idiosyncratic behavior to accomplish this that makes them not resemble normal ES6 classes, and behave in ways that may be surprising or unclear to users who aren't familiar with Alt patterns.

I'd like for actions to both offer a terse, low-boilerplate syntax, while still following the lines of more standard ES6 objects, with more intuitive behavior for this.

I propose to have 3 categories of properties on an `Actions` object;
1. Standard actions that just forward their payload
2. Actions that transform and forward their payload
3. Normal methods that aren't actions at all (but may serve as action creators that dispatch other actions)

I think the code should look like this:

``` js
class FooActions extends Actions {
  static actions = ['simpleAction', 'getFooStarting', 'getFooDone', 'getFooFailed'];
  // Or maybe this.generateActions(...) in constructor.

  @action
  actionWithTransform(value) {
    return this.getTransformed(value);
  }

  getTransformed(value) {
    return 2 * value;
  }

  getFoo(id) {
    this.getFooStarting({id});
    FooWebApiUtils.getFoo(id).then(
      result => this.getFooDone({id, result}),
      error => this.getFooFailed({id, error})
    );
  }
}
```

What you end up with is that `FooActions.simpleAction` is just a simple action that will dispatch its payload, and that `FooActions.actionWithTransform` is an action that will apply a simple transform to its payload before dispatching it.

However, `FooActions.getTransformed` isn't an action at all, and will just return the transformed value. Likewise, `FooActions.getFoo` is not an action - it's a standard Flux-y action creator that dispatches other actions (by calling their methods).

My concerns with this proposal are:
1. Is it sufficiently obvious that e.g. above, `getFooDone` is an action but `getFoo` is an action creator? Per my earlier proposal, maybe use `CONSTANT_CASE` for actions and `camelCase` for regular methods?
2. This is proposal is incompatible with some Alt patterns - if you have an action that looks like e.g. `foo(value) { fooFuture().then(() => this.dispatch(value)); }`, you would need to create a new e.g. `fooDone` action and dispatch that (via `this.fooDone(value)`), as `fooDone`, rather than as `foo` as it would be right now. This is the flip side of having a firm separation between actions and action creator or helper methods.

taion:
[...]

PR #503, opened 2015-10-03
Remove this.dispatch from actions

### LINKED ISSUE #502: README: shouldn't updateTodo be called addTodo?
Seems like this is just adding the todo to the list.  It doesn't actually update anything, right?

```js
class TodoStore {
  ...
  updateTodo(todo) {
    this.setState({ todos: this.state.todos.concat(todo) });
  }
}
```

Or am I missing something important about how alt works?

goatslacker:
This is a large breaking change, fortunately a codemod is included here to automatically transform your code into the working future. 

This change has the benefit of making Action classes actual classes where you can store things on `this` or call other actions using `this.action()` rather than having that weird `actions` object.

Fixes #502 and #359 

The only caveat from #359 is that you don't get actions that don't dispatch and instead transform some data or whatever. This isn't a deal breaker for me (you can always define an outer function to transform data) but not sure how everyone else feels.

Alternatively this allows us to introduce syntax such as `this.preventDispatch()` along with `extends Actions` which will allow you to have methods in an action class that don't dispatch.


taion:
I don't think this actually fixes #501. If I try

```js
class Actions {
  constructor() {
    this.foo = 2;
  }
  one() {
    return 1
  }
  two() {
    console.log('foo', this.foo)
    return 2
  }
}
```

in `foo.js`, calling `actions.two()` logs `foo undefined`.

More broadly I think this example actually pushes against having action classes at all. In the above case, I don't see the benefit of using an ES6 class there rather than just doing (assuming it worked)

```js
actions = {
  foo: 2,
  one() {
    return 1
  },
  two() {
    console.log('foo', this.foo)
    return 2
  }
}
```

mehcode:
```
actions = {
  foo: 2,
  one() {
    return 1
  },
  two() {
    console.log('foo', this.foo)
    return 2
  }
}
```

---

```
  },
  two() {
```

It's the `,`. It's always been the `,`.

taion:
ಠ_ಠ

goatslacker:
Aside from the `,` having a class lets you call things on `this`. Let me check this out because that should work fine since the tests are passing.

taion:
@goatslacker Wouldn't it be best if it were possible to call things on `this` for both class objects and POJSOs?

taion:
On second thought, at least when using instances rather than singletons, it seems like it'd just make more sense to use classes rather than POJSOs for actions, maybe.

goatslacker:
Ok fixed.

taion:
Cool - so this means that for both classes and POJSOs, non-functions will show up as-is, while functions will all get turned into actions that dispatch their return values (unless they return a promise)?

One other edge case - what's the right behavior if the actions class extends another class?

goatslacker:
instance properties work, actions don't currently.

taion:
Looks great to me BTW.

goatslacker:
Let me rebase and make sure that I don't need to do anything else before I land this.

goatslacker:
Also, prior to releasing 0.18.0 I wonder if I should also figure out any other breaking changes to bundle with this one, or make upgrading more digestible by introducing only this breaking change.

taion:
People on long enough upgrade cycles are going to end up having to do multiple upgrades at once anyway... I've had people bugging me about 5 month old installs of React-Bootstrap and trying to upgrade to React v0.14 without upgrading R-B.

Is there anything big in the pipeline? I guess maybe the server-side data fetching stuff?

goatslacker:
Nah I'm abandoning that, leaving it up to userland.

taion:
Huh I feel like that needs a bit of integration at the container level to work well... but I guess those containers can be in userland too.

I don't have any opinion on the stores thing.

ziflex:
I would add my 2 cents - what if Alt have DI?
Now we can easily get any store and action from alt instance using 'getActions' and 'getStore', but what if we could register any custom services and call 'getService' for it ?

PR #510, opened 2015-10-13
Add check for Object.isFrozen() in isPojo()

tony99nyr:
This adds support for seamless-immutable.js which freezes property fields. Without this change StateFunctions.js will attempt to delete frozen keys and throws an error. This simple additional check for Object.isFrozen() enables alt.bootstrap({}) to work with seamless-immutable and potentially other immutable libraries which freeze property fields.


tony99nyr:
@goatslacker Please take a look at this PR. Incredibly simple change that is blocking me. Thanks!

goatslacker:
Can you add a test for this? I'm curious that this doesn't work out of the box, a frozen object is just an object after all.

tony99nyr:
True, but in the uses of isPojo() I have seen, the object checked is immediately modified. Thus in these context "Plain Old Object" assumes it can make changes.  

Ultimately, it is the following deletion attempt  in `utils/StateFunctions.js` which is blocking me.  If it makes sense to put the freeze check somewhere else I really don't mind.

``
if (fn.isPojo(state)) {
     fn.eachObject(k => delete state[k], [state])
     ...
}
``

Thanks for your time.

goatslacker:
ah fair enough, we should rename this then from isPojo to isModifiableObject or something.

tony99nyr:
Have you run the `'frozen store state can be bootstrapped'` test I added? It is a concise example of the problem I am facing.  Let me know if you would like me to make any further changes.

goatslacker:
Yeah totally it's legit. Last change: can we rename the isPojo function to something like isMutableObject?

tony99nyr:
Done. isMutableObject() is perfect, thanks.

goatslacker:
I'll cherry pick this onto 0.17.6 and release it as 0.17.7

tony99nyr:
Perfect, thanks man.

On Thu, Oct 22, 2015, 4:42 PM Josh Perez <notifications@github.com> wrote:

> I'll cherry pick this onto 0.17.6 and release it as 0.17.7
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/goatslacker/alt/pull/510#issuecomment-150350470>.
>
-- 

Thanks,
Tony Rossi

tony99nyr:
I saw that you merged into master thank you. Could you please do an `npm version patch` and `npm publish` pretty please?

Thanks!

goatslacker:
Published as alt@0.17.7

PR #520, opened 2015-10-22
Update link from boilerplate to generator

talyssonoc:
I've created a generator for that boilerplate as I find it easier for people to use and it kind of _deprecated_ the boilerplate (that will be used for experimentation now, not as a base for developing).


goatslacker:
Thanks :)

PR #523, opened 2015-10-29
Update broken AltContainer link

azmenak:


goatslacker:
Ah thanks. All the links need updating actually.

PR #538, opened 2015-11-09
The use of 'props' needs explaination

ethan-deng:
The use of props is never explained in the doc and the code is incorrect.


goatslacker:
Thank you!

PR #543, opened 2015-11-14
Match babel v6

MattMcFarland:
Spent the better part of 12 hours stuck and this was the culprit.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "stage": 0
+  "presets": ["stage-0"]
 }


MattMcFarland:
while alt may be using babel 5. I'm wokring on a project that is using babel 6 and jest with react.  Jest and babel-jest would NOT work until I changed your .babelrc file.  Sucks that it took me over 12 hours to figure it out.  Not sure if you want to support v5 or v6, but it might be a good idea to at least have a separate version for v6.  It's really babel 6 that's the culprit, but I already have invested in it with other stuff and this turned out to be the easiest fix.

taion:
Why are you transpiling `node_modules` in the first place? You shouldn't be.

@goatslacker FWIW my preferred solution (https://github.com/rackt/react-router/pull/2479, https://github.com/rackt/history/pull/137) is to whitelist `files` instead of using `.npmignore` so you don't accidentally ship `.babelrc`... but that really only helps users who misconfigure things.

MattMcFarland:
I'm not transpiling node_modules.  I am using babel-jest, which I guess does?? I'm not sure.. but I'm not touching node_modules, I never do.  Only in rare occassions when there is a bug and of course why I shipped out a PR.  

Are you saying we could actually remove the .babelrc??

taion:
In general there's no good reason to distribute your `.babelrc`.

Can you have babel-jest exclude `node_modules`?

taion:
By "distribute" I mean they shouldn't be in the npm package. I don't think this PR is correct though.

MattMcFarland:
No, babel-jest has to include the module 'alt'  I believe, but I can try without it..  You can see in alt examples it pulls in the alt module as well:  https://github.com/goatslacker/alt/blob/master/examples/chat/package.json#L36

Am going to try without it..  Thanks.

FWIW I dont think this is a good PR either, I think if we can, preventing the .babelrc from being in here in the first place would be best.

MattMcFarland:
FYI: I had removed the entry to alt in my package.json and was able to revert the changes back.  Thanks for your responses.  I'm not sure why the example has it.  I can say that it's really tough getting jest to work with alt, but that might just be my lack of intimate knowledge of alt itself.

THanks again

taion:
No worries - do you want to put together a PR to either update `.npmignore`, or switch to a whitelist (assuming @goatslacker is okay with that)?

MattMcFarland:
Yes.

taion:
Great, thank you - looking forward to it.

goatslacker:
I'll publish a new version. 

Alt probably won't move to babel6 soon until all the bugs are sorted out.

taion:
@goatslacker How are you planning on dealing w/stuff like `'alt/utils/connectToStores'` for CJS users after switching to Babel 6?

MattMcFarland:
Happy to help.  I think one of the issues is that there may be a use-case for not mocking alt, when using jest.  I cannot think of one right now, but I have not gotten really deep into unit testing, actually I've just barely scratched the surface.   I'm not really sure about Jest anymore.  In theory it seemed awesome, but after trying to get it to work all night I'm beginning to think I might use something else.  What do you guys use for unit testing alt + react?  Do you have any pointers?

goatslacker:
@taion connectToStores no longer exists in the alt npm package

taion:
Ahh okay but either way. The whole changing the default export thing.

ljharb:
@taion my reason for distributing babelrc is the same as for distributing the source ES6 - so that if someone wanted to review the original code, they'd have the original transpilation settings, which can vary widely across projects.

taion:
That's fine in principle but in practice ends up causing pain for users that forget to suppress transpilation on `node_modules`, especially given mutual choking by Babel 5 and 6 on the respective opposite version config files.

MattMcFarland:
Since babelrc is not in .gitignore, then you are free to distribute the
bablerc file with the source.  However I dont see the reason to include it
with npm.. unless there is a good reason for that as well?

On Wed, Dec 9, 2015 at 4:53 PM, Jimmy Jia <notifications@github.com> wrote:

> That's fine in principle but in practice ends up causing pain for users
> that forget to suppress transpilation on node_modules, especially given
> mutual choking by Babel 5 and 6 on the respective opposite version config
> files.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/goatslacker/alt/pull/543#issuecomment-163406107>.
>



-- 

Matt McFarland

Software Engineer | @docodemore <https://twitter.com/docodemore> | LinkedIn
<http://www.linkedin.com/pub/matthew-mcfarland/6a/b9a/135/en> | Mobile:
513.638.8685 | Office: 937.865.6800 xt: 125010

ljharb:
@taion users who forget to suppress transpilation of `node_modules` have already caused themselves pain. the sooner that chokes the better. if this is a webpack flaw, then it's something webpack needs to fix, not "every single module publisher that uses babel".

@MattMcFarland typically the source is included alongside the compiled output, for those that want to review it. If the source *is* included, I'm claiming that the babelrc should also be included. Certainly if you do not include the source, babelrc shouldn't be included.

taion:
@ljharb 

1. That's admirable in principle, but Babel 5 and Babel 6 `.babelrc`s are not compatible - in practice if people did that, nobody would be able to ship anything until the ecosystem converged.

2. It's very rare to ship the source on npm, at least in the React ecosystem. Consult e.g. `react` itself.

PR #544, opened 2015-11-14
remove babelrc

MattMcFarland:
Because there are different versions that contain different schema for a .babelrc file, it's probably best to avoid keeping this file in version control.


goatslacker:
Thanks

PR #570, opened 2015-12-15
Update Fetching Data docs for 0.18

sloria:
- Remove `this.dispatch` in favor of returning a thunk.
- `this.actions.doSomething()` -> `this.doSomething()`

PR #572, opened 2015-12-15
Remove this.dispatch usage from docs

sloria:
Together with #570 and #571, all usages of `this.dispatch` in the docs should be updated to use the new 0.18 syntax.

### DIFF
--- a/docs/testing/actions.md
+++ b/docs/testing/actions.md
@@ -39,15 +39,17 @@ class PetActions {
         // adds import charge to cost
         totalCost = importCost + cost;
 
-    this.dispatch({
-      pet,
-      cost: totalCost
-    });
-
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -846,7 +846,7 @@ var action = {
 
 var AppDispatcher = Object.assign(new Dispatcher(), {
   handleAction(payload) {
-    this.dispatch(payload);
+    return payload;
```
goatslacker: this is actually vanilla flux

goatslacker: can you please revert this. everything else is :+1:

sloria: Ah, careless mistake. Fixed and pushed.

goatslacker:
Thanks!

PR #574, opened 2015-12-16
Update createActions docs

goatslacker:
Ref #571 

to @sloria


goatslacker:
I was thinking of keeping this stuff in createActions because it might still be useful.

sloria:
Looks good!

PR #577, opened 2015-12-17
Bringing back this.alt to actions

m7moud:
This is my attempt on bringing back this.alt to actions, not much of a hack, but I kept the `exportObj` as it is to make the change as minimal as possible.


m7moud:
I removed `exportObj` from the `createActions` function signature because it will break `addActions` down if you passed something else to it instead of an empty object, since `addActions` expects just an object of actions.

CrazyPython:
Wait, why should an action need access to the alt instance? Shouldn't actions be pure functions?

m7moud:
:D This was three years ago, I don't even remember why I create this PR.

PR #588, opened 2016-01-07
Remove broken example links from README

wbreakell:
This removes a pair of broken links to examples that no longer exist in this repository. These examples were removed in this commit https://github.com/goatslacker/alt/commit/f365116db93b899f92c676eebfc21a167655c5d1


wbreakell:
Hmm, I'm not sure why the build's failing.

jdlehman:
Before we remove the links, I want to make sure they have not moved elsewhere (because then we can link to the other repos). For reference: [commit they were removed](https://github.com/goatslacker/alt/commit/f365116db93b899f92c676eebfc21a167655c5d1)

@goatslacker, do we plan on moving these examples to `altjs` org?

goatslacker:
Yeah we can move these over to that repo. For now we can remove them from the README.

PR #599, opened 2016-01-17
Updating documentation on connectToStores alt util

christemple:
I noticed that since installing 0.18.1 that the utils have now been removed.

The documentation has fallen behind a little with regards to the `View` section when it comes `connectToStores` so I have updated to let users know to install the `alt-utils` package and get connectToStores from there.

Also, when I was trying to use alt.js for the first time I was having problems with the fact the View example was using the ES7 `@connectToStore` decorator which really threw me off as I hadn't looked at any ES7 features and had no idea what to search for to find why it wasn't working for me.  So I updated it to include ES6 and ES7 examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,14 +67,16 @@ class TodoStore {
 export default alt.createStore(TodoStore, 'TodoStore');
 ```
 
-View
+View 
+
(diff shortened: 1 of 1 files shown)


jdlehman:
Looks good, thanks for making these updates!

PR #600, opened 2016-01-18
Linking to bootstrap doc page from takeSnapshot

ldabiralai:
I thought this might be useful for newcomers to alt, as it shows the relation between the two methods.

Open to suggestions on how to improve this though.


jdlehman:
Thanks

PR #606, opened 2016-01-25
Fix failing Promises example from the documentation.

### LINKED ISSUE #584: Documentation lists non-working Promise example Actions
this code: 
````alt.createActions({
  fetchUser(id) {
    return function (dispatch) {
      dispatch(id);
      return http.get('/api/users/' + id)
        .then(this.fetchUserSuccess.bind(this))
        .catch(this.fetchUserFailure.bind(this));
    };
  },

  fetchUserSuccess: x => x,
  fetchUserFailure: x => x,
});

alt.actions.fetchUser(123).then(() => {
  renderToHtml(alt); // we know user fetching has completed successfully, and it's safe to render the app
});
````

Will flat out not work. What will happen is that you'll get an error thrown saying `alt.actions.fetchUser.then is not a function` which of course it isn't.

What's actually returned is the anon function that's passed dispatch. This makes using promises AND dispatching the action outright impossible right now (as far as I can tell from the documentation anyhow). It appears that `this.dispatch` was removed (also spread around in the examples and docs) so there's no way to dispatch the action if you return a promise correctly.

ie - this works for `alt.actions.fetchUser(123).then(()`:

````
fetchUser(id) {
    return http.get('/api/users/' + id)
        .then(this.fetchUserSuccess.bind(this))
        .catch(this.fetchUserFailure.bind(this));
  }
````

But the action will never be dispatched. This is doubly troublesome for Server Side rendering. 

Thoughts? Currently it seems that promise support has been kind of butchered. I'm assuming it's due to some roadmap changes...

mareksuscak:
This PR fixes #584.

### DIFF
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -18,7 +18,7 @@ export default function makeAction(alt, namespace, name, implementation, obj) {
     // async functions that return promises should not be dispatched
     if (invocationResult !== undefined && !isPromise(invocationResult)) {
       if (fn.isFunction(invocationResult)) {
-        // but should return an inner function result
+        // inner function result should be returned as an action result
         actionResult = invocationResult(dispatch, alt)
       } else {
         dispatch(invocationResult)
--- a/src/actions/index.js
+++ b/src/actions/index.js
@@ -12,22 +12,23 @@ export default function makeAction(alt, namespace, name, implementation, obj) {
 
   // the action itself
   const action = (...args) => {
-    const result = implementation.apply(obj, args)
+    const invocationResult = implementation.apply(obj, args)
+    let actionResult = invocationResult
 
     // async functions that return promises should not be dispatched
-    if (result !== undefined && !isPromise(result)) {
-      if (fn.isFunction(result)) {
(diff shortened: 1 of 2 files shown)


mareksuscak:
@goatslacker Can you also consider merging this PR?

goatslacker:
Yup I'll take a look

goatslacker:
@mareksuscak can you link to the example in the documentation.

Also, should we still dispatch on success?

mareksuscak:
@goatslacker Sure, here it is: http://alt.js.org/docs/createActions/#dispatching - example at the bottom. Not sure what you mean by dispatching on success. This PR modifies the async action invocation code which will return an enclosing (inner) function's return value.

goatslacker:
Ah I get it now. Thanks :D

goatslacker:
Will publish this.

mareksuscak:
Awesome. Thanks. Looking forward to new release.

goatslacker:
Published as 0.18.2. Thanks again!

PR #607, opened 2016-01-25
Fix dev deps so that the build doesn`t break after the initial setup.

mareksuscak:
After checking out the source code for the first time and `npm install` the error was being displayed:

screen shot 2016-01-25 at 20 10 00

The reason was a missing `babel-loader` and `style-loader` dependency. This should be gone with this PR merged in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "babel-cli": "6.3.17",
     "babel-core": "6.3.17",
     "babel-eslint": "5.0.0-beta6",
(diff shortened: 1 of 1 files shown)


mareksuscak:
Build timed out. That's the reason why it failed. I am not allowed to trigger a rebuild and don't want to send more unnecessary commits in here so please trigger a rebuild whoever can do it.

goatslacker:
Thanks

PR #610, opened 2016-01-27
Remove reference to this.dispatch

lsanwick:
`this.dispatch` no longer exists, having it in the docs is confusing.


goatslacker:
Thanks!

PR #615, opened 2016-02-05
Removed dispatch from typings

pdolega:
Apparently new Alt version ( 0.18.x ) removed `dispatch` from Actions.


pdolega:
Weird, I seem to be unable to build it... It fails on Travis.

goatslacker:
Thanks!

PR #632, opened 2016-03-17
return data from remote function call in source, allows for data to be returned in then on promise

kellyrmilligan:
This allows for a component to call a source method(on the store), and if chaining a then to the promise, receive the data from the promises resolve. all tests pass and coverage stays at 100%. Also updated babel-eslint to be out of beta to fix linting.


kellyrmilligan:
?

goatslacker:
2731395e7c61ea45c7320badf1de0e58f8bf10c9 Thank you!

PR #636, opened 2016-04-05
Updated link to boilerplate

wassname:
Updated link to boilerplate as react-webpack-node now has their alt implementation on a branch.


jdlehman:
Thanks

PR #645, opened 2016-04-26
Fix state after returning new state object from reduce.

JorgenEvens:
After returning a new state object from a reduce method there are now two state objects being used by the store. The one stored on `model.state` and the one stored on the `AltStore` instance.

The easiest way to see this is by checking whether the state passed into the `reduce` method is the same as that available on`this` ( the model as reduce is called on it ). I would expect it to be, but it is not.


JorgenEvens:
I've added node 5.11 to the list of node engines to test on Travis-CI as Babel does not seem to be to happy with node v6.

PR #669, opened 2016-07-22
Future of alt

goatslacker:
Hi! This is just here for your perusal. 

At some point I'd like to merge this in but it needs a lot of documentation and I'd like to clean up package.json.

Please voice your concerns with this new direction in this PR.

If you're interested in checking out the code, I recommend you give this branch a try. You can also read the code over in `src/` which is now just three files and it's a lot simpler.

Here's a snippet:

``` js
import Alt, { Store } from 'alt';

const todoActions = alt.generateActions('todoActions', ['addTodo'])

class TodoStore extends Store {
  constructor() {
    super()
    this.state = { todos: [] }
    this.bindActions(todoActions)
  }

  addTodo(todo) {
    this.setState({ todos: this.state.todos.concat(todo) })
  }
}

const todoStore = alt.createStore('todoStore', new TodoStore())
```


avdeveloper:
will there be a documentation update along with these new changes?

goatslacker:
Yes. All of the docs and examples. Before merging this I'd like to revamp all of the docs.

andrejbaran:
Hey @goatslacker,

First of all, thanks for all your effort put into alt, it's really appreciated.

Second, I want to point out an entry in the changelog:

> +15. no more promises being returned from actions

That is kind of in _conflict_ with how `createAsyncActions` works:

``` js
  createAsyncActions(namespace, actions) {
    return Object.keys(actions).reduce((obj, actionName) => {
      const type = `${namespace}/${actionName}`

      const dispatch = (payload, opt) => {
        const meta = {}
        if (opt.loading) meta.loading = true
        const action = { type, payload, meta }
        if (opt.error) action.error = true

        this.publish(action)
        return action
      }

      obj[actionName] = (...args) => {
        const payload = actions[actionName](...args)

        dispatch(null, { loading: true })

        return Promise.resolve(payload).then(
          result => dispatch(result, {}),
          error => dispatch(error, { error: true })
        )
      }
      Object.assign(obj[actionName], { type, actionName })
      return obj
    }, {})
  }
```

`payload` would still be a Promise, wouldn't it? Otherwise it's not really an async action.

Third, at company I work for we decided to go with alt over other tools mostly because of the it's design. For example having the ability to use classes (es6) for actions/stores, having multiple stores and others was exactly what we needed and were looking for. I understand the need to take alt further but we'd love to see this feature retained.

Furthermore, there's #684 that completes the ability to use classes for actions. Do you think there's a chance this would make it's way into 1.x?

ryanzec:
@goatslacker I don't have any concerns but I do have a question (and gitter seems pretty dead so I will ask here).

This line:

> no more sources

Does this mean the whole Data Sources stuff (http://alt.js.org/docs/async/)?

I am not saying this is bad, personally I find the data sources stuff with the way it is implement a little limiting, I'm just asking for clarification (and if that it what it is, it should probably be `no more data sources` so it is clear to other as I am sure I am not the only one with this question).

icoloma:
Maybe actions created under the same namespace should be merged instead of replaced. This would make it possible to mix different ways of creating actions: 

```
// generated actions
alt.generateActions('LocationActions', [ 'foo', 'bar' ]);

// synchronous actions
alt.createActions('LocationActions', LocationActions);

// async actions
alt.createAsyncActions('LocationActions', LocationAsyncActions);
```

geoffRGWilliams:
@ryanzec it does indeed mean the removal of data sources... if it ever comes to that. Confirmed by @goatslacker.

ryanzec:
@geoffRGWilliams yea, at this point, I have considered alt pretty much dead and the version we have now is the "future" of alt (ie. there will be no new versions).

PR #684, opened 2016-09-13
getPrototypeChain gets whole prototype chain

andrejbaran:
Hi,

this changes `AltUtils.getPrototypeChain` to return methods from the whole prototype chain (instead of just from the first prototype), which we needed since we have multiple subclasses in the chain for the final `ActionClass`.

I'm not sure you guys want something like that merged, but in case you do here's the patch.

Thanks for the feedback


andrejbaran:
Just a hint regarding the tests, you have fairly old versions of babel and babel plugins and tests are failing because of that. I can add a commit that fixes that if you want me to. Test were passing on my local setup with updated versions.

PR #695, opened 2016-10-30
SCREAMING_SNAKE_CASE action

ntwcklng:
Hi, 

it would be nice if we inform new users about this pattern in the getting started section.


ntwcklng:
ping @goatslacker

PR #736, opened 2019-11-27
Fix listen dispose/unlisten memory leak

davidosomething:


src/store/AltStore.js
```diff
@@ -89,19 +89,26 @@ class AltStore {
...
   }
 
   unlisten(cb) {
     this.lifecycle('unlisten')
-    this.subscriptions
-      .filter(subscription => subscription.cb === cb)
-      .forEach(subscription => subscription.dispose())
+    this.subscriptions = this.subscriptions.filter((s) => {
+      if (s.cb === cb) {
+        subscription.dispose()
```
jpoo90: @davidosomething I think you want `s.dispose()` here.

davidosomething: fixed!

